PR #4, opened 2019-01-03
Fixed bug which was causing an unwanted "a" key event.

pouwelsjochem:
This bug happened when using the screenshot tool in MacOS. Simply handled it by just ignoring the event when the keyCode = kVK_ANSI_A because this shouldn't even happen in the 'flagsChanged' method.

Steps to reproduce the original bug: Simply add a "key" listener in the main.lua and print the keyName & phase. Start the simulator and create a screenshot with "cmd + shift + 4". Runtime:addEventListener("key", function(event) print(event.keyName, event.phase) end)


Shchvova:
Thanks a lot! You submitted first ever community fix. Congrats and much gratitude!

I also noticed that window doesn't get key-up events, but I guess it's not much can be done with it.

PR #7, opened 2019-01-05
Improved display.newSprite(), 'time' can now be an array of frame-times.

pouwelsjochem:
Previously you were only able to pass a number as 'time' parameter, which limited sprites to just have constant frame times. For example `{name="default", frames = {1,2,3,4}, time = 400, loopCount = 1}` would result in a sprite with 100ms per frame.

By allowing an array of frame-times, you could make sprites with variable frame-times. For example
`{name="default", frames = {1,2,3,4}, time = {50, 120, 80, 150}, loopCount = 1}`


librtt/Display/Rtt_SpriteObject.cpp
```diff
@@ -100,31 +126,59 @@ SpriteObjectSequence::Create( Rtt_Allocator *allocator, lua_State *L, int index
...
+	lua_getfield( L, index, "time" );
+	if ( lua_isnumber(L, -1) )
+	{
+		time = luaL_toreal( L, -1 );
+	}
+	else if ( lua_istable( L, -1 ) )
+	{
+		numFramesInTimeArray += (int) lua_objlen( L, -1 );
+		timeArray = (Real *)Rtt_MALLOC( allocator, numFramesInTimeArray * sizeof( Real ) );
+		for ( int i = 0; i < numFramesInTimeArray; i++ )
```
Shchvova: like in here, do `for ( int i = 0; i < numFramesInTimeArray && i<numFrames; i++ )`
And later something like
```
for ( int i = numFramesInTimeArray;  i<numFrames; i++ )
    timeArray[i] = lastGoodValue
```
And do static check and trace if numFramesInTimeArrat!=numFrames

librtt/Display/Rtt_SpriteObject.cpp
```diff
@@ -302,6 +354,26 @@ SpriteObjectSequence::GetFrame( int index ) const
...
+	{
+		return frameIndex * GetTimePerFrame();
+	}
+	else
+	{
+		Real summedTime = 0;
+		Real *timeArray = GetTimeArray();
+		for (int i = 0; i < frameIndex; ++i)
+		{
+			summedTime += timeArray[i];
```
Shchvova: It seems like we never actually use `timeArray[i]` directly. We always use time of array start. So may be instead of storing `timeArray[i]` we can summarize it right away. Like, instead have a `frameEndTimeArray` where `frameEndTimeArray[i] = frameEndTimeArray[i-1] + timeArray[i]`, or something.

This way it seems we can skip ever iterating the `timeArray` or store cached `fTimeArrayCached*` values. I'm not sure, but it seems we can skip storing `timeArray` at all this way.

librtt/Display/Rtt_SpriteObject.cpp
pouwelsjochem: Makes sense, latest commit changed the SpriteObjectSequence::Create implementation to only allocate the timeArray once.

librtt/Display/Rtt_SpriteObject.cpp
```diff
@@ -58,9 +58,39 @@ SpriteObjectSequence::DirectionForString( const char *value )
...
-SpriteObjectSequence*
-SpriteObjectSequence::Create( Rtt_Allocator *allocator, lua_State *L, int index )
+Real*
+SpriteObjectSequence::VerifyValidTimeArrayParam( Real *timeArray, int numFramesInTimeArray, int numFrames )
 {
+	if (numFramesInTimeArray == 0 || numFramesInTimeArray == numFrames)
+	{
+		return timeArray;
+	}
+	else if (numFramesInTimeArray > numFrames) // Crop
```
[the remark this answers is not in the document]
pouwelsjochem: Makes sense, latest commit changed the SpriteObjectSequence::Create implementation to only allocate the timeArray once.

librtt/Display/Rtt_SpriteObject.cpp
pouwelsjochem: I'm not a fan of not storing timeArray at all, since doing that would make SpriteObject::GetFrameIndexForDeltaTime more complex (and slightly less performant).

We could store an additional frameEndTimeArray or a cache which would remember the expected result for SpriteObjectSequence::GetTimeForFrame. But it sounds like premature optimising to me. Especially since SpriteObjectSequence::GetTimeForFrame is only executed when calling sprite:setFrame(), which I think would not be used a lot on sprites with variable frame times.

Issue #13, opened 2019-01-21
Adding the ability to handle 'characters' instead of just 'keys'


pouwelsjochem:
At the moment Corona is able to listen to "key" events from all of the supported OS's. This works perfectly fine when for example it is being used for controlling your in-game character. But doesn't work well when you're trying to catch actual keyboard input from the user.

The problem is that there's only a `keyCode` property, but no mapping to the device-independent character which takes modifier keys and keyboard layout (AZERTY, QWERTY, QWERTZ, etc..) in consideration. All of the supported OS's can do this in native code though. MacOS & Windows support this in their KeyUp/Down events, but Linux and HTML would require listening to an additional event type which triggers when a printable character is pressed.

I'd suggest implementing a new simple `character` event with the following properties. 
`name`: which would always resolve to "character"
`device`: which would resolve to an InputDevice object
`character`: which would resolve to the localised character, taking keyboard layout and modifier keys in consideration.

PR #14, opened 2019-01-23
Added the CharacterEvent for MacOS, Linux and Windows

pouwelsjochem:
Added the CharacterEvent for MacOS, Linux and Windows. Solves issue https://github.com/coronalabs/corona/issues/13


platform/windows/Corona.Native.Library.Win32/Interop/Input/Key.cpp
```diff
@@ -362,6 +363,38 @@ const char* Key::GetCoronaName() const
...
+		switch (ToUnicode(this->GetNativeCodeValue(), scanCode, keyboardState, chars, BUFFER_LENGTH - 1, 0))
+		{
+		case -1:
+		case 0:
+			return NULL;
+		default:
+			WinString stringConverter;
+			stringConverter.SetUTF16(chars);
+			int utf8Length = strlen(stringConverter.GetUTF8()) + 1;
+			char * out = new char[utf8Length];
```
Shchvova: This is a memory leak it seems. Probably should wrap it in std::string or something

PR #16, opened 2019-02-11
Add timeTransform to graphics.defineEffect

ggcrunchy:
This is an opt-in mechanism aimed at mitigating numeric precision woes in shaders that result after the timer has run for a while.

Three techniques are provided. I considered adding noise as well, such as the 1D version [here](https://web.archive.org/web/20080724063449/http://freespace.virgin.net/hugo.elias/models/m_perlin.htm). Other ideas welcome.

The following code gives a side-by-side comparison of the methods:

```
graphics.defineEffect{
    category = "generator", group = "time_tests", name = "none",

    isTimeDependent = true,

    fragment = [[
        P_COLOR vec4 FragmentKernel (P_UV vec2 _)
        {
            return vec4(CoronaTotalTime, 0., 0., 1.);
        }
    ]]
}

local r1 = display.newRect(100, 100, 50, 50)

r1.fill.effect = "generator.time_tests.none"

graphics.defineEffect{
    category = "generator", group = "time_tests", name = "modulo",

    isTimeDependent = true, timeTransform = { func = "modulo", range = 5 },

    fragment = [[
        P_COLOR vec4 FragmentKernel (P_UV vec2 _)
        {
            return vec4(CoronaTotalTime / 5., 0., 0., 1.);
        }
    ]]
}

local r2 = display.newRect(200, 100, 50, 50)

r2.fill.effect = "generator.time_tests.modulo"

graphics.defineEffect{
    category = "generator", group = "time_tests", name = "pingpong",

    isTimeDependent = true, timeTransform = { func = "pingpong", range = 5 },

    fragment = [[
        P_COLOR vec4 FragmentKernel (P_UV vec2 _)
        {
            return vec4(0., CoronaTotalTime / 5., 0., 1.);
        }
    ]]
}

local r3 = display.newRect(300, 100, 50, 50)

r3.fill.effect = "generator.time_tests.pingpong"

graphics.defineEffect{
    category = "generator", group = "time_tests", name = "sine",

    isTimeDependent = true, timeTransform = { func = "sine", amplitude = .5, period = 5 },

    fragment = [[
        P_COLOR vec4 FragmentKernel (P_UV vec2 _)
        {
            return vec4(0., 0., CoronaTotalTime + .5, 1.);
        }
    ]]
}

local r4 = display.newRect(400, 100, 50, 50)

r4.fill.effect = "generator.time_tests.sine"
```

### DIFF
--- a/librtt/Display/Rtt_ShaderFactory.cpp
+++ b/librtt/Display/Rtt_ShaderFactory.cpp
@@ -554,9 +554,11 @@ PingPong( Real *x, Real range, Real, Real )
 }
 
 static void
-Sine( Real *x, Real amplitude, Real period, Real shift )
+Sine( Real *x, Real amplitude, Real speed, Real shift )
 {
-	*x = amplitude * Rtt_RealSin( period * (*x + shift) );
+	Real t = *x;
+
+	*x = amplitude * Rtt_RealSin( speed * t + shift );
 }
 
 static void
@@ -636,7 +638,7 @@ ShaderFactory::BindTimeTransform(lua_State *L, int index, const SharedPtr< Shade
 
 				transform = Rtt_NEW( fAllocator, TimeTransform );
 
-				transform->func = &Modulo;
+				transform->func = &Sine;
 				transform->arg1 = amplitude;
 				transform->arg2 = (Rtt_REAL_2 * M_PI) / period;
 				transform->arg3 = shift;
--- a/librtt/Display/Rtt_ShaderFactory.cpp
+++ b/librtt/Display/Rtt_ShaderFactory.cpp
@@ -667,7 +667,10 @@ ShaderFactory::InitializeBindings( lua_State *L, int shaderIndex, const SharedPt
 	ShaderData *defaultData = Rtt_NEW( fOwner.GetAllocator(), ShaderData( resource ) );
 	resource->SetDefaultData( defaultData );
 
-	BindTimeTransform( L, shaderIndex, resource );
+	if (resource->UsesTime())
+	{
+		BindTimeTransform( L, shaderIndex, resource );
+	}
 
 	bool has_vertex_data = BindVertexDataMap( L, shaderIndex, resource );
 	if( has_vertex_data )
--- a/librtt/Display/Rtt_ShaderFactory.cpp
+++ b/librtt/Display/Rtt_ShaderFactory.cpp
@@ -533,6 +533,106 @@ ShaderFactory::BindUniformDataMap( lua_State *L, int index, const SharedPtr< Sha
 	return result;
 }
 
+// STEVE CHANGE
+static void
+Modulo( Real *x, Real range, Real, Real )
+{
+	*x = fmod( *x, range ); // TODO: Rtt_RealFmod?
+}
+
+static void
+PingPong( Real *x, Real range, Real, Real )
+{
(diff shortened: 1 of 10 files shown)


ggcrunchy:
**Some commentary / motivation, for docs:**

Corona's shaders use [IEEE-754 floats](https://en.wikipedia.org/wiki/IEEE_754) as the underlying representation for their numeric values.

For the majority of cases (the exceptions are irrelevant here), part of the number constitutes an integer numerator. Let's call this _N_. Our numerator can go from 0 to _D_ - 1, _D_ being a constant power of two. Together these give us a scale factor _t_ = _N_ / _D_ in the range [0, 1).

The rest of the number is devoted to the sign (positive or negative) and the exponent, the latter being an integer that gives us a power of 2, for instance 2<sup>5</sup>. We obtain values by interpolating between neighboring powers, using the scale factor: _result_ = 2<sup>p</sup> * (1 + _t_) Notice that if _t_ were 1 we would land on the next power of 2.

This can exactly represent some values, but will only approximate most. Any format is going to have such idiosyncracies. IEEE-754 happens to provide a lot of accuracy near 0 while also giving us exact integers all the way up to 2 * _D_.

Lua offers us 64-bit floats, with rather generous 53-bit numerators and accompanying accuracy. On GPUs we are rarely so lucky, especially with mobile hardware, owing to concerns like bandwidth and memory.

For instance, see the ["Qualifiers"] (https://www.khronos.org/opengles/sdk/docs/reference_cards/OpenGL-ES-2_0-Reference-card.pdf) section in this OpenGL ES 2.0 reference card. With **mediump** our _D_ is only guaranteed to be an underwhelming 1024.

Now imagine what this means for time, measured in seconds. At first, we'll be just fine. But just after the two-minute mark, interpolating between 128 and 256, we only go in steps of (256 - 128) / 1024, or 1/8 second accuracy. At five minutes we'll proceed in increments of 1/4, and so on. Anything relying on such results becomes quite choppy.

This gloomy scenario only captures part of what's going on, of course. The time is actually maintained in Lua as a single-precision float, with a respectable 23-bit numerator. Furthermore, many shaders can get by with rather small values derived from the time, for example one built on a trigonometric function that only needs some low multiple of &pi;. In these cases the shader can take a transformed result, such as `TrueTotalTime % X` or `sin(N * TrueTotalTime)`.

**And usage**:

If `isTimeDependent` is defined, Corona will also look for `timeTransform`. This must be a table and should have one of the following as its *func* member: "modulo", "pingpong", "sine". The value of `CoronaTotalTime` within the shader will be the result of any such transformation.

The "modulo" transform is computed as `CoronaTotalTime = CoronaTotalTime % range`, where `range` is a positive number that may be supplied in the `timeTransform` table under that same key. By default, `range` is 1.

The "pingpong" transform is similar, except `CoronaTotalTime` will first go from 0 to `range`, then fall back to 0, then repeat.

The "sine" transform is computed as `CoronaTotalTime = amplitude * sin(scale * CoronaTotalTime + phase)`. Again, `amplitude` and `phase` may be provided in the `timeTransform` table, with defaults 1 and 0 respectively. The scale is calculated from a `period` parameter, a positive number indicating how much time should pass before the sine waves repeats. The default is
2 * &pi;, which has a scale factor of 1.

ggcrunchy:
Hmm, it looks like I'm unconditionally looking up the transform stuff, even though it should probably be ignored when `isTimeDependent` isn't true. I think the `BindTimeTransform()` call would be guarded by a `if ( resource->usesTime() )`, but will wait on this for now.

PR #17, opened 2019-02-24
add xDelta & yDelta properties to touch events

schroederapps:
This commit modifies "touch" events so that the event table dispatched now includes "xDelta" and "yDelta" properties, indicating the difference between the touch's current X/Y values and the "xStart" and "yStart" values.

Added as a convenience so that developers won't have to perform these calculations in their touch listeners.

More-or-less copied from the undocumented Runtime "drag" event.


schroederapps:
(Sorry I'm seeing the incorrect indentation here in GitHub - but that is not present for me in Xcode where I made the edits.)

Shchvova:
You can easily add commits to the pull request to fix indents.

schroederapps:
Fixed indents. Sorry 'bout that!

tabs not spaces

PR #18, opened 2019-03-05
Render state objects

ggcrunchy:
These are non-drawing, non-hit testable "display" objects that, if "visible", will affect objects that follow in the group, then (in most cases) be cleaned up in reverse order at the end. At the moment most of these toggle or temporarily set state not exposed by display object properties, often stuff that makes more sense across multiple objects, e.g. color masks, stencil settings, etc.

One of these "state" changes is also a stencil clear command, so the groundwork is laid for one-off actions as well, although the mechanisms might need to mature some.

This opens up a few features, e.g. stencils are handy for outlines. It also opens up some plugin possibilities, such as [NanoVG](https://github.com/memononen/nanovg), which depends on them.

Other features, such as various depth buffer-related things, are stubbed out but not implemented. (Docs yet to come.)

An example / test:

```
local g1 = display.newGroup()
local g2 = display.newGroup()

local aa = display.newRect(g1, 100, 100, 200, 200)
local bb = display.newRect(g1, 600, 200, 200, 100)

aa:setFillColor(0, 1, 0)
bb:setFillColor(0, 0, 1)

local rr = display.newRenderState(g1)

rr.colorMask = { blue = false }

local cc = display.newRect(g1, 350, 200, 100, 150) -- will be white minus blue = yellow

local dd = display.newRect(g2, 250, 400, 300, 100)

dd:setFillColor(1, 0, 1) -- will be red

local ss = display.newRenderState(g2) -- scissor out anything that follows in g2

ss.scissorEnabled = true
ss.scissor = { x = 300, y = 300, width = 250, height = 250 }

for i = 1, 20 do
  local c = display.newCircle(g2, math.random(250, 600), math.random(250, 600), math.random(6, 25))

  c:setFillColor(math.random(), math.random(), math.random())
end

local xx = display.newRect(550, 300, 100, 100)

xx:setFillColor(0, 1, 1) -- unaffected, color mask and scissor cleaned up at end of g1 and g2 respectively

local was = display.getDefault("stencilBitsize")

display.setDefault("stencilBitsize", 8) -- add stencil to snapshots or canvas textures...

local s = display.newSnapshot(300, 300)

display.setDefault("stencilBitsize", was) -- ..and disable

s.x = display.contentCenterX
s.y = display.contentCenterY

local tt= display.newRenderState()

s.group:insert(tt) -- enable stencil write of 1

tt.stencilEnabled = true
tt.stencilFunc = { ref = 1 } -- defaults: func = "always", mask = 0xFFFFFFFF
tt.stencilOp = { depthPass = "replace" } -- defaults: stencilFail = depthFail = "keep"
tt.colorMask = { green = false, alpha = false } -- defaults: red = blue = true

local r = display.newRect(20, 20, 100, 100)

s.group:insert(r)

r:setFillColor(0, 1, 0) -- no green will be written, but color only masked; stencil WILL be written

local uu= display.newRenderState()

s.group:insert(uu)

uu.stencilFunc = { func = "notEqual", ref = 1 } -- forgo parts of stencil we already covered
uu.stencilMask = 0 -- no values written, even on the outline
uu.colorMask = { alpha = true } -- turn alpha back on for rest of objects in snapshot

local t = display.newRect(20, 20, 100, 100)

s.group:insert(t)

t:scale(1.1, 1.1) -- extend the object so we still etch an outline
t:setFillColor(1, 0, 0)
```

This introduces a few files and I haven't been able to dip into non-Windows projects yet, so a lot of the logic is behind guards along with an intentionally annoying comment for later search:

```
#if Rtt_WIN_ENV // TODO: RenderStateObject REMOVE ME!
   // stuff
#endif
```

These were first compiled as `#if 0` for verification.

I added a line in the Windows setup to provide stencil buffers. I don't know if other platforms will require the same.


Shchvova:
For now I am closing this pull request. I hope we'll integrate this later.

PR #32, opened 2019-12-14
Added initialSystemUiVisibility property to 'android' key in build.settings

pouwelsjochem:
This sets the systemUiVisibility even before the splashScreen comes up, avoiding the resize of the corona logo (which normally happens when settings immersiveSticky in main.lua)
For Android devices < 4.4 setting a initialSystemUiVisibility with `immersiveSticky` will automatically fall back to `lowProfile` mode.

Also added View.SYSTEM_UI_FLAG_LAYOUT_HIDE_NAVIGATION flag to immersive & immersiveSticky modes.

This will prevent the resize event on Android which happens when starting the application. The application will immediately have the right dimension.
Without this, a Corona app on a 1920x1080 screen will start at 1776x1080 and resizes to 1920x1080 shortly after starting up (a "resize" event occurs)
With this, a Corona app will immediately start with 1920x1080 as display.pixelWidth & display.pixelHeight


sekodev:
@Shchvova I haven't seen this referenced in the documentation. Is it something that should be added to the docs?

PR #33, opened 2019-12-21
Added initialSystemUiVisibility to the valid_build_settings.lua. Also…

pouwelsjochem:
… used proper indentation in update_manifest.lua


Shchvova:
This change will be in next daily build, but I will not start a build for it.

PR #35, opened 2020-01-20
Added new native.setProperty() named "windowSize" for MacOS

pouwelsjochem:
This will allow you to change the window width/height in code by passing a table with either `width` or/and `height` values


Shchvova:
@pouwelsjochem 2020.3567 is up.

PR #47, opened 2020-03-30
Allow Android to handle repeated volume button events

pouwelsjochem:
This will keep the functionality of not sending repeated "key" events to lua. But allows Android to decrease/increase the volume.


Shchvova:
It's been a while. Why do you think this should work like that?

pouwelsjochem:
Android sends the volume buttons at an interval, currently the implementation blocks the system from handling repeated events by returning true. This change would return false on repeated volume buttons, thus enabling Android to handle them natively.

Tested it on Android 8 or 9, not sure which one it was exactly

PR #49, opened 2020-04-19
Removed failed to download plugin alert

scottrules44:
Removed failed to download plugin alert
not needed because there is a print statement


Shchvova:
Thanks for doing this. Probably best solution for now. I'm working in my own for on a varsion which would support stubs from custom urls. Would be ready before the 1st of May.

PR #66, opened 2020-05-10
Allow to change the timeScale of an active SpriteObject

depilz:
If you set the time scale of an sprite object that is currently playing, it will keep the frame and allow smooth time scale transitions


Shchvova:
Good job!

PR #72, opened 2020-05-19
isDistributionBuild should be true when there are no provisioned devices

joehinkle11:
I'm fairly certain this solves the problem, but I haven't been bothered to clone the whole repo on my machine and build this to check. I figured out this flag was backwards because every time I built with my built with my dist cert I got an .app, which was mildly infuriating. AND CoronaBuilder would say "Build type: 'distribution'" when I built with my dev cert and "Build type: 'developer'" when I built with my dist cert.

### DIFF
--- a/tools/CoronaBuilder/Rtt_AppPackagerMacFactory.mm
+++ b/tools/CoronaBuilder/Rtt_AppPackagerMacFactory.mm
@@ -114,7 +114,7 @@ AppPackagerFactory::CreatePackagerParamsInternal(
 
 			NSString *provisionFile = [NSString stringWithUTF8String:certificatePath];
 
-			bool isDistributionBuild = [AppleSigningIdentityController hasProvisionedDevices:provisionFile];
+			bool isDistributionBuild = ![AppleSigningIdentityController hasProvisionedDevices:provisionFile];
 
 			IOSAppPackager packager( fServices );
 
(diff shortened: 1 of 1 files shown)


joehinkle11:
@Shchvova Do you think you could merge this when you get the chance? I'd like to use new versions of Solar with this fix for CoronaBuilder

Shchvova:
I just tried your fix, it doesn't do anything. If you want to test it, I suggest here are simple steps to follow:
1. Clone repo `git clone --recursive https://github.com/coronalabs/corona.git`
2. Copy templates from the current installation of Solar2D into the `corona/platform/resources/iostemplate`:
```
 cp -r /Applications/Corona-3596/Corona\ Simulator.app/Contents/Resources/iostemplate/*.tar.bz corona/platform/resources/iostemplate
```
3. Open the CoronaBuilder project: `open corona/platform/mac/CoronaBuilder.xcodeproj`
4. Build this targets: `lua`, `luac`, `CoronaBuilder` and `rttplayer`.
5. Start the last target. Now you can actually build, and test your code.

joehinkle11:
troylyndon from the forums seems to have solved this problem. 

> At the end of your provision profile, I see it is for “Apple Distribution” NOT “iOS Distribution”. FYI, although Apple has designed a way to use a single profile for all it’s platforms, I haven’t found a way to use it. To solve this, just make sure to choose iOS Distribution in all cases.

https://forums.solar2d.com/t/issue-building-for-ios-distribution/351543/2?u=joecoronasdk

PR #75, opened 2020-05-23
Update README.md

Destiny-2048:
Minor changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # Solar2D Game Engine
-## rebranded Corona CoronaSDK or Corona
+## Rebranded CoronaSDK or Corona
 > Simple to learn and use, completely free and open source 2D game engine.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
 # Solar2D Game Engine
-## rebranded Corona CoronaSDK or Corona
+## Rebranded Corona SDK or Corona
```
Veter51, line 2: Maybe Corona SDK with a space would be better

Destiny-2048, line 2: Maybe Corona SDK with a space would be better

sekodev:
I think just "Rebranded Corona SDK" is far better. IMO, repeating the same name is not entirely necessary.

(By the way, it could be a good idea to sign CLA to submit your changes.)

README.md
Shchvova: Thanks. I agree

PR #85, opened 2020-06-07
Init SSL Library before calling ctx new.

agramonte:
SSL Library not initialized before calling SSL CTX new. Resulting in a null CTX.

At some point, the SSL use to be initialized someplace else (maybe with the license check). In any case, currently, the SSL is not initialized before the call to create a new CTX. This results in a null CTX that gets passed to another method and crashes the live server.


Shchvova:
I wonder why I commented it out...

PR #114, opened 2020-07-18
Adding support easy integration with solar2d marketplace

scottrules44:
I have done my testing on the rttplayer with Xcode
If you want to test it you can add this to your build.settings
plugins = {
		["plugin.firebaseAnalytics"] =
		{
			publisherId="tech.scotth",
			marketplaceId = "jdt0u5"
		},
	}

and this to your main.lua

local firebaseAnalytics = require ("plugin.firebaseAnalytics")
firebaseAnalytics.init()

This is not the best example because this plugin is available on plugins.solar2d but you get the idea.
If the user wants to use the solar2d marketplace they can choose to do so by just adding marketplaceId.

I hope to add better error support in the future. 

Thanks,
Scott

### DIFF
--- a/platform/resources/CoronaBuilderPluginCollector.lua
+++ b/platform/resources/CoronaBuilderPluginCollector.lua
@@ -171,8 +171,14 @@ function PluginCollectorSolar2DDirectory:collect(destination, plugin, pluginTabl
         return params.canSkip or "Solar2D Directory: skipped plugin " .. plugin .. " because platform " .. pluginPlatform .. " is not supported"
     end
     local repoName = pluginObject.p or (pluginTable.publisherId .. '-' .. plugin)
-    local downloadURL = "https://github.com/" .. repoOwner .. "/" .. repoName .. "/releases/download/" .. pluginObject.r .. "/" .. vFoundBuildName .. "-" .. pluginPlatform .. ".tgz"
+    local downloadURL
 
+    if pluginTable.marketplaceId then
+      downloadURL = "https://solar2dmarketplace.com/marketplacePlugins?pluginType=collector&ID=" .. pluginTable.marketplaceId .. "&plugin=" .. plugin .. "_" .. pluginTable.publisherId .. "&type=" .. pluginPlatform
+    else
+      downloadURL = "https://github.com/" .. repoOwner .. "/" .. repoName .. "/releases/download/" .. pluginObject.r .. "/" .. vFoundBuildName .. "-" .. pluginPlatform .. ".tgz"
+    end
+	
     local cacheDir = pathJoin(params.pluginStorage, "Caches", "Solar2Directory", repoOwner, pluginTable.publisherId, plugin, pluginPlatform )
     local cacheUrlFile = pathJoin(cacheDir, "info.txt")
     local cacheDestFile = pathJoin(cacheDir, "data.tgz")


Shchvova:
I'm not sure how this suppose to work. But I can see some problems with code:
1. Solar2D relies on plugins being unchanged. Basically, all versions are numbered with releases, so it won't download same release twice. 
2. Solar2D Directory is not connected with marketplace. So plugins have to be both in free plugin directory and in your marketplace for this to work.

I suggest making a new collector for your marketplace which would account for that. You can start by copying Solar2DPluginCollector.

scottrules44:
> I'm not sure how this suppose to work. But I can see some problems with code:
> 
> 1. Solar2D relies on plugins being unchanged. Basically, all versions are numbered with releases, so it won't download same release twice.
> 2. Solar2D Directory is not connected with marketplace. So plugins have to be both in free plugin directory and in your marketplace for this to work.
> 
> I suggest making a new collector for your marketplace which would account for that. You can start by copying Solar2DPluginCollector.

Just to explain a little better, I want to make it so if a user adds there marketplace account id under the plugin settings for example 
["plugin.firebaseAnalytics"] =
{
publisherId="tech.scotth",
marketplaceId = "jdt0u5"
},

The marketplaceId is account Id from the solar2d marketplace. When a user adds marketplaceId this indicates that user want to get a plugin from solar2dmarketplace.com, not from solar2d free plugins.

1. Could I get around this by adding this? 
local validCache = false
    if isFile(cacheUrlFile) and isFile(cacheDestFile) then
        local f = io.open(cacheUrlFile, "rb")
        if f then
            validCache = f:read("*all") == downloadURL
            f:close()
        end
    end

    if pluginTable.marketplaceId then
      validCache = false
    end

I could also try to add version support on my server to handle the caching.


2. I would honestly be okay with adding all plugins to the 
https://plugins.solar2d.com/plugins.json and handle it like the class kit plugin.
In the future, I could make my own JSON file or maybe you could expose a way for me to submit a plugin to this directory programmatically.


Do you suggest making a copy of CoronaBuilderPluginCollector.lua or just adding on to CoronaBuilderPluginCollector?
It feels unnecessary to make my own copy CoronaBuilderPluginCollector but if that's what you recommend.

Shchvova:
Sorry it took so long to get around for this reply. So, in the file, take a look at the object
`PluginCollectorSolar2DDirectory`
You can copy it and create your own plugin collector. It can be object or a function. If it is an object, it should have `collect(destination, plugin, pluginTable, pluginPlatform, params)` function.
It returns string in case of error, or true if plugin is found and put in place.
Goal is to place data.tgz into the `destination` directory. Collector would be called multiple times, even for same `plugin`. For example if plugin not found for `win32` it would look for same plugin in `win32-sim` and then in `lua` as `pluginPlatform`.
You can access `pluginTable`, to check out publisherId, etc.

Also, you don't have to insert it into the Core. You can develop plugin collector separately `~/Solar2DPlugins/Collectors` on macOS. All .lua files from there will be added to collector list, just make sure to return the collector object (or a function).

Plugin collectors have many convenience functions accessible to them. Most notably `fetch` and `download` for retrieving web locations as strings and downloading files. I suggest reading contents of the `CoronaBuilderPluginCollector.lua`. I tried to make it plain.

Also, collector object has `init` method, called only once. Idea is to get list of authorized plugins there or do similar initialization to avoid making network calls for every plugin.

I think ideal solution would be to generate plugin collector for user with their authorization keys embedded into in, asking them to put into directory and everything is working.

PR #128, opened 2020-08-17
Adds a lateUpdate event called after the enterFrame event

depilz:


Shchvova:
What is the reasoning for having it? Like, how it is different from [enterFrame](https://docs.coronalabs.com/api/event/enterFrame/index.html) event?

depilz:
It follows the same idea as the [Unity lateUpdate](https://docs.unity3d.com/ScriptReference/MonoBehaviour.LateUpdate.html)

Some guys were interested in an event that is always triggered after all the other enterFrames and also, just before the game renders.

From what I understood, the enterFrame is literally the last process before the game starts rendering, so this will be useful in both cases.

There is a short discussion about this in the #general channel on discord.

Shchvova:
Let’s give it a try why don’t we.

sekodev:
@depilz don't forget to add this to documentation, please. Otherwise, most people won't notice this brand new event type.

I don't know how many references in documentation for `enterFrame` but here are the ones I found at first glance:
https://docs.coronalabs.com/api/event/index.html
https://docs.coronalabs.com/guide/events/detectEvents/index.html

depilz:
@sekodev I'm working on that right now. @Shchvova can you confirm if this late update is going to be last call before the game renders?

Next to the `display -> update()` (Where the lateUpdate() is) there is a `display->Render()`. This seems pretty clear to me that there is the only point where the game starts rendering, but maybe there are more things I don't know about.

superqix:
Sorry, late to the party on this, but it's a welcome update and it mirrors some Unity functionality that's pretty interesting. Nice work all!

depilz:
Thanks @superqix ! And thanks for the update @sekodev . The PR for the new lateUpdate event is [here](https://github.com/coronalabs/corona-docs/pull/49)

PR #138, opened 2020-08-31
iOS: initial support of Metal rendering backend (early Alpha)

Shchvova:


kakashidinho:
I just noticed Solar2D creates and uploads luminance texture data every frame (for displaying text it seems). Luminance texture format data upload is currently not very efficient in MetalANGLE due to conversion from luminance to RGBA. I will try to optimize it.
However, there is a better way. You could change `Texture::kLuminance` handling code in `getFormatTokens()` (https://github.com/coronalabs/corona/blob/fb1dd283ca52d0ace430fe7470aaab24c320dcbd/librtt/Renderer/Rtt_GLTexture.cpp#L39) to use `GL_RED_EXT` instead `GL_LUMINANCE` (`GL_RED_EXT` is from extension `GL_EXT_texture_rg` but MetalANGLE supports it on all Metal devices)

kakashidinho:
Running #139 sample code on simulator I saw noticeable micro stuttering due to format conversion. The stuttering went away after changing `GL_LUMINANCE ` to `GL_RED_EXT`.

Shchvova:
Thanks so much. I had to cut some corners to make it work at first, but results are awesome so far.
Also, if you noticed I changed to my fork, which uses submodules and some minor changes. Would you like me to make a pull request for them?

kakashidinho:
Thanks, the submodules changes I don't want to merge just yet. I still want to keep the old dependencies structure so that it would be easy to merge with upstream ANGLE repo.

I saw you changed min iOS SDK to 8.0. That's nice but I'm afraid it would introduce some compatibility issues that I'm not aware. So let's see how it goes with Solar2D project for now.

PR #146, opened 2020-09-08
Add a UINavigationController

Larpoux:
Hi, everybody,

I need to switch the Current View from Solar2D, to a private native View :
```
             FlutterViewController* flutterViewController = [[FlutterViewController alloc]  
                        initWithEngine:flutterEngine nibName:nil bundle:nil];
             [nav pushViewController: flutterViewController animated:YES];
```
        
And, then, back to Solar2D : 
```
UINavigationController* nav = titi.navigationController;
[nav popViewControllerAnimated: YES];
```
 
I was not able to run correctly this code until I patched the procedure `initializeWithApplication()` in `AppDelegate.mm` : 

```
      // [LARPOUX] [ESTRELA] needs a Navigation Controller
        UINavigationController* navigationController = [[UINavigationController alloc]initWithRootViewController: viewController];
        window.rootViewController = navigationController;
        [window setRootViewController: navigationController];
        // Get warnings in iOS 6 if we don't do this.
        // [LARPOUX] [ESTRELA] // [window setRootViewController:viewController];
```
I am not an iOS expert but this patch does not seem to disturb the correct working of S2D.

- Is there anybody who know how to avoid this patch ?
- May I submit a Pull Request with this patch ?

Thank you, guys, and girls


Shchvova:
Sorry. What is the reason for this change? Does it fix or improve something?

Larpoux:
Hi Vlad,
I need to push my own View from  Solar2d OpenGL View, and after pop my View to return to Solar2d.
To do that, I need a Root View Controller.

I think (and test) that there is no negative effect on Solar2d if you create a View Controller during initialization.

Larpoux:
A navigation controller, sorry, not a View controller.

Shchvova:
I just worry it may affect on plugins which rely on the current hierarchy. Would you like me to help with your actual code? PM me on [discord](https://discord.gg/WMtCemc), maybe we’ll solve your issue in the other way?

Larpoux:
Sure.

PR #168, opened 2020-10-01
Mask canvas fix

ggcrunchy:
Hi. This is an update of [this PR](https://github.com/coronalabs/corona/pull/29). On some platforms there was nothing visible; this affects at least Android and web. I haven't had a chance to try the latte, though the fix there might be as simple as adding the corresponding `defined(x)`.

As mentioned [here](https://stackoverflow.com/questions/14718322/opengl-es-cant-make-luminance-fbo-work-in-android/18687435#18687435), luminance isn't a legal render target on GL ES ([similarly](https://stackoverflow.com/a/54444344) seems to suggest WebGL will be similar). `GL_RED_EXT` would probably be more appropriate than RGBA, but that calls for an extension query step that I haven't done yet.


Shchvova:
Looks good.

PR #170, opened 2020-10-05
Update CoronaBuilderPluginCollector.lua

scottrules44:
Made Solar2dmarketplace collector get the newest version if available


Shchvova:
This is great!
Did you consider submitting missing stubs to S2D Free plugin directory, so users would know where to find your paid plugins?
You submitted two of your plugins already. Here’s Lerg’s pull request to do it with his plugins: https://github.com/solar2d/plugins.solar2d.com/pull/10
Only thing it does is customizes the failure message if plugin is not found, so users can know where to look for it.

PR #172, opened 2020-10-08
Android: restoring android.apkFiles functionality

Shchvova:


agramonte:
Is this functionality documented? When would you want to include apk files onto a resource.

PR #175, opened 2020-10-10
Update gradle to 6.5 for JDK 14

daisy1754:
Gradle 6.1 has issue with JDK 14 https://github.com/gradle/gradle/issues/10248
Updating to 6.5 so people can build with JDK 14


Shchvova:
This is irrelevant now, since Java is bundled into the distribution. However, updating gradle is a good idea anyway.

PR #193, opened 2020-11-12
Per-vertex fill and stroke colors

ggcrunchy:
This lets us assign fill / stroke colors, for shape and line objects, to each vertex. (These will be scaled by the "true" fill / stroke colors, if both are in effect.) This began life as an API for meshes, where we already have some per-vertex control; there was only a very thin layer between mesh shapes and the underlying shape path objects, so it was actually a bit cleaner to open it up.

Most objects' vertex layout differs a little from what the user sees, so the respective vertex counts are provided. Even without knowing the fine layout details we can still do a few things, e.g. setting random vertex colors as in this sample:

[perVertex.zip](https://github.com/coronalabs/corona/files/5533374/perVertex.zip)

All told we have:

```
  local fillVertexCount = object.fillVertexCount -- shapes
  local strokeVertexCount = object.strokeVertexCount -- shapes and lines
 
  object:setFillVertexColor(index, r, g, b) -- shapes; index from 1 to fillVertexCount; colors can take forms like setFillColor()
  object:setStrokeVertexColor(index, r, a) -- shapes and lines
```

The sample shows various shapes, plus an image, with some mix of color assignments, mostly random. Some helper functions (set fill colors, set stroke colors, set both) are doing the bulk of the work and can easily be swapped out. The rect also has a demonstration of swapping a "true" fill color in and out while also having per-vertex fill colors.

The implementation adds color arrays that are lazily allocated when the first valid vertex color assignment is performed. The vertex counts will use the allocated values when possible, but these will be briefly invalid when the object is just created, or when points have been added to a line; a dummy tessellation is performed in these situations and the resulting vertex count used. (In lines this might be overzealous since modifying the stroke width and so on could also trigger the invalidation.)


Shchvova:
First of all... This is awesome!

librtt/Display/Rtt_ShapePath.cpp
```diff
@@ -410,6 +410,58 @@ ShapePath::GetTextureExtents( const ArrayVertex2& texVertices ) const
 	return extents;
 }
 
+bool
+ShapePath::SetFillVertexColor( U32 index, U32 color )
+{
+	fFillGeometry->AttachPerVertexColors( &fFillSource.Colors(), GetFillVertexCount() );
```
Shchvova, line 416: I'm not sure I like this path... This pretty much ensures that tessellation happens twice always. Do we need to `GetFillVertexCount` here always? Do you think there's a way to attach vertex colors without knowing number of vertices in advance, then just ignoring "extra" colors?

Shchvova: Basically, it is fine to ask to tesselate if we explicitly request number of vertices. But if we know it in advance, we can skip tessellation, like for mesh or quad. Not sure it is an issue, as it would have to happen eventually anyways, we're just removing laziness.
what do you think? I'm not asking to make changes here (yet) just for a discussion.

ggcrunchy: Thanks!

I would love a better solution. In my original implementation with meshes only, I was able to ask the (mesh-specific) tessellator for its `Length()`. For most of the other shapes and lines, adding this probably wouldn't be so hard, as say a `virtual`. With circles (and maybe rounded rects) I wasn't quite as confident, or at least expected some pain. 😄

ggcrunchy: Whoops, saw you already mentioned meshes.

In retrying my sample, I see that path really does get hit hard, especially with the circle! I'm looking over the tessellators and it doesn't look too bad to add a vertex count method. In particular, all the strokes will be (# points + 1) * 2. Even those rounded ones don't look so bad, now.

I'll take a stab at this.

ggcrunchy:
Okay, I traced my way through the fill and stroke count and seem to now be successfully computing most of the counts. I compromised with the polygon fill count and the mesh stroke count: calling the tessellation logic but saving the result and reusing that if not -1.

Shchvova:
Good enough 😆
This is awesome. Thanks so much

zero-meta:
Can I set perVertexColor with alpha channel?
The code below doesn't work as expected. It should be a gradient transparent color.
But it's not transparent at all.
``` lua
local mesh = display.newMesh({
    x = 20,
    y = 20,
    mode = "indexed",
    vertices = {
        0,0, 0,500, 500,500, 500,0
    },
    indices = {
    	1, 4, 2,
    	2, 4, 3
    }
})
mesh:translate(mesh.path:getVertexOffset())
mesh:setFillVertexColor(1, 1, 1, 1, 0.1)
mesh:setFillVertexColor(2, 1, 1, 1, 0.9)
mesh:setFillVertexColor(3, 1, 1, 1, 0.9)
mesh:setFillVertexColor(4, 1, 1, 1, 0.1)
```
image

ggcrunchy:
The mixing happens around here: <https://github.com/coronalabs/corona/blob/master/librtt/Display/Rtt_Paint.cpp#L203>

I guess I was a bit careless about premultiplied alpha. Will need to review that. 😄

ggcrunchy:
@zero-meta Are you familiar with building the engine? In the last few days I've been playing with per-vertex alpha ([here](https://github.com/ggcrunchy/fur-texture/blob/master/geometry1.lua#L327) if you're interested), so far just adding

```
if (isPremultiplied)
{
	vc.rgba.PremultiplyAlpha();
}
```

after the `vc.pixel = perVertexColors[i];` line.

I'll try to throw some simpler tests together&mdash;just for rects and their corners, say&mdash;but if you have an opportunity you could try the above change or something similar. (I'll still need to test mixing a fill and per-vertex fill with alpha, at least.)

PR #230, opened 2021-02-21
Update iPhonePackageApp.lua

scottrules44:
Turns out the "catch" was the on demand resources was not being bundled in with IPA itself for iOS. Added line a couple lines during build process to move the resources into the ipa if onDemandResources is in build.settings


Shchvova:
I'll try testing with TestFlight tomorrow

scottrules44:
I already tested this in Testflight but the more eyes the better :)

PR #243, opened 2021-04-17
Add optional parent and position parameters to the display.newGroup()

depilz:
With this you can create groups using display.newGroup([parent], [x, y])


XeduR:
Given that the old/current optional parameters are undocumented and referred to as experimental, and as no one in Discord seemed to know about them, I think this change would be a fantastic change and a welcome addition.

depilz:
Any chance of accepting this some day?

PR #256, opened 2021-06-08
Android: Google Play Asset Delivery initial support

Shchvova:
When  `android.usesExpansionFile = true` all assets would be placed into the preloaded asset pack. This will expand game size to 1GB.

### DIFF
--- a/platform/android/.gitignore
+++ b/platform/android/.gitignore
@@ -83,3 +83,5 @@ lint/tmp/
 android.ipr
 android.iws
 /sdk/.cxx/
+
+preloadedAssets/src/
--- a/platform/android/app/build.gradle.kts
+++ b/platform/android/app/build.gradle.kts
@@ -17,7 +17,7 @@ val coronaResourcesDir: String? by project
 val coronaDstDir: String? by project
 val coronaTmpDir: String? by project
 val coronaAppFileName: String? by project
-val coronaAppPackage = project.findProperty("coronaAppPackage") as? String ?: "com.corona.test"
+val coronaAppPackage = project.findProperty("coronaAppPackage") as? String ?: "com.corona.app"
 val coronaKeystore: String? by project
 val coronaKeystorePassword: String? by project
 val coronaKeyAlias: String? by project
@@ -244,6 +244,9 @@ android {
     }
     aaptOptions {
         additionalParameters("--extra-packages", extraPackages.filter { it.isNotBlank() }.joinToString(":"))
+    }
+    if (isExpansionFileRequired) {
+        assetPacks.add(":preloadedAssets")
     }
     // This is dirty hack because Android Assets refuse to copy assets which start with . or _
     if (!isExpansionFileRequired) {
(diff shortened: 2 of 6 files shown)


xbcross:
Hello, if I need to use this function, do I need to add code into my project manually? Or does corona already have this function?

Shchvova:
@xbcross Solar2D handles everything. You only need to add the setting

xbcross:
Excuse me, I've seen your example of solar2d. The Android project does not include this modification. The user needs to configure it himself, right? Please let me know!

xbcross:
@Shchvova Excuse me, I've seen your example of solar2d. The Android project does not include this modification. The user needs to configure it himself, right? Please let me know! thank you

XeduR:
@xbcross https://docs.coronalabs.com/guide/distribution/buildSettings/index.html#expansion-files

xbcross:
@Shchvova Excuse me, after I configured according to your document and generated "AAB", I tried to use Google's bundletools tool for local testing, but unfortunately it failed. I don't know what the problem is. Can you put a demo for my reference? Because I don't see any trace of this modification in the example of "corona-3648 / national / samples / simpleluaextension / Android". So I don't know where my problem is. Hope to get your help, thank you!

Shchvova:
@xbcross sorry. All setup you have to do is to set the value in your `build.settings` file:
```lua
{
    android = {
        usesExpansionFile = true
        ...
    }
}
```
That's that. No other setup is required (or possible).
Then make sure to select Google as target store when building the app.
To test the packaging you can unzip the produced `aab` file. In the output you will get additional `preloadedAssets` directory. You can unzip with 7z on windows or The Unarchiver on macOS.
(also, instead just APK you will get apk and oob files)
There are no samples for it yet. All what changes is literally the single line in `build.settings`. You can not influence it with any other code in your app.

I tested it only with Google Play uploading the AAB to use in closed alpha.

xbcross:
@Shchvova OK, thank you very much. I misunderstood it. Thank you again for your help. Thank you

xbcross:
@Shchvova Hello, I have encountered a problem again. I try to build my project with "Android studio". The generated "aab" has no "preloadedassets" folder after it is unwrapped. Did you build it directly with solar2d? If I need to build it with "Android studio", what should I do? Thank you very much for your

Shchvova:
Is there a reason you're not using the Simulator? This flag only works when packaging with the Sim. With Android Studio you would have to make your own asset packs. It isn't complicated but I can help you with that if you really need to use Android Studio.

xbcross:
@Shchvova Yes, I need Android studio to build my project. Thank you for your help

xbcross:
@Shchvova How do I need to make my asset package to build with Android studio? I'm still confused. I hope you can help me. I made a similar production according to your source code, but I didn't find the "preloadedassets" file after unpacking the constructed "aab". If you can help me, I'm really happy

PaoUNetWork:
@Shchvova Excuse me, I also encounter the same problem. My project also needs to be built with ‘Androidstudio’. What can I do to build with ’Androidstudio‘? I have the same question. Can you directly provide a ‘demo project ’for your reference! Thank you here!

PR #280, opened 2021-09-25
Allow local file access in Android webview

prairiewest:
Required since API level 30, as they changed the default from True to False.

https://developer.android.com/reference/android/webkit/WebSettings#setAllowFileAccess(boolean)


Shchvova:
You can use setNativePropery for this. Do it right after creating the WebView.

PR #284, opened 2021-10-06
Mask canvas fix, generalized

ggcrunchy:
I first got this working on Windows, then learned it was failing at least on Android, and eventually fixed that.

I've recently had access to a Mac again, and saw that it was also broken elsewhere.

This generalizes the Android fix to all non-Windows platforms.


ggcrunchy:
If [this post](https://forums.solar2d.com/t/need-help-about-texture-mask/354392/2) is accurate, even Windows might need the fix, to work in general.

Detecting (and implementing) "R" format support would probably be the best way to go, but that's another round of work. :smile:

ggcrunchy:
I can confirm that the issue now happens even on one of my Windows machines; I don't quite know what update or whatever changed it, but I suppose it _is_ more conformant that way. Without the `#ifdef` / `#endif` it works fine.

PR #293, opened 2021-10-20
Custom objects, step #2: custom commands

ggcrunchy:
This is **PR 2** as first described [here](https://github.com/coronalabs/corona/pull/283).

---

There is slight overlap with **PR 1** with the `Rtt_ApplePlatform.mm` and `Main.cpp` files, since more library functions have been added. The remaining files are all distinct from **PR 1**; no new ones this time.

That said, this can't just be built on its own, since it depends on the public types and object box list stuff from **PR 1**. It's just an attempt to keep the overall process digestible. 😄 

---

The details:

Solar's display hierarchy is not responsible for the actual rendering, but instead structures display objects into consecutive geometry batches of related state, then issues a series of commands, including `Draw`s. The "command buffer" is where rendering finally happens.

To do custom, display hierarchy-based rendering, we need to issue such commands. The custom objects from **PR 1** ensure that we are ready and waiting at the appropriate moment, while the new features here give us the command capability.

---

Again, the comments should give a reasonable approximation of the eventual docs.

---

I tested with [this plugin](https://github.com/ggcrunchy/solar2d-plugins/tree/master/engine_testing/custom-objects-test2). Using the machinery from **PR 1**, we create a color mask object and put it in the group before any children we want to affect. The object's `Draw()` method issues a custom Solar command that calls `glColorMask()` at the appropriate time.

A special "Do" routine is used to perform the aforesaid logic. This ensures that Solar first commits any pending non-masked display objects.

Since it's in a "scope group" (also used in the **PR 1** plugin), the mask object will receive a **"didDraw"** message when the last member of the group is drawn, i.e. when the "scope" ends. This will rewind to the mask state to what it was before. (If I wrote this properly you can stack masks + scope groups, although this might not be a very important use case. :smile:)

When a mask command is issued, an "end of frame" handler is temporarily installed. This is a failsafe in case some mask object wasn't in a scope group: we explicitly restore a non-masked state so the next frame begins as expected.

Although not used in this plugin, similar support is available for "clear" handlers, for resources&mdash;stencil or depth buffers, say&mdash;that should be wiped before any rendering.


ggcrunchy:
In working on the **PR 5** stuff (WIP [here](https://github.com/ggcrunchy/corona/wiki/WIPs)), I realized the "Do" approach and the discontinuities it introduced into the renderer batch, just didn't cut it. You can now reserve a blob of dirtyable state that is examined, and reacted to, in `Insert()`. This removes the need for "Do" as well as the end of frame handlers (or rather, these are subsumed into one for the dirty state).

The clear handler was also done away with, albeit in a different way (basically, the "important" cases were added to `system.getInfo()` / `display.setDefault()`.

I figure this won't suddenly be pulled, so just pushed the updated plugins that rely on some of that **PR 5** machinery.

Shchvova (approved):

PR #299, opened 2021-11-11
Feat/meshes

depilz:
Added a new `mesh.path:update(params)`, which is used as follows:

```lua
mesh.path:update{
   vertices = params.vertices,
   uvs      = params.uvs,
   indices = params.indices
}
```

I tested it with many Spine animations and everything is working as expected and with a performance gain of 30-70% depending of the animation, these are some samples:

CleanShot 2021-11-11 at 18 33 22

CleanShot 2021-11-11 at 18 33 55


depilz:
I also updated the Spine version 3.7, 3.8 and 4.0 to work with this, not approved yet of course, but I can probably share those files thought here or Discord.

Shchvova:
This is awesome. Shall merge ASAP

PR #338, opened 2022-01-24
Moving linux build to snapcraft

Shchvova:


bakeinflash:
Let's try to commit it. Looks like very good. It should work.

PR #345, opened 2022-02-16
Adding 'android-samsung' as a plugin type.

agramonte:
Adding it so I can create a specific version of Hyperbid for Samsung devices.


Shchvova:
May be it would be better idea to add "non-google" meaning android without play services and such? But this works too, imo.

agramonte:
Actually, that is a much better idea. Please reject this one and I'll change it to that.

PR #360, opened 2022-03-19
Linux development

bakeinflash:
- removed dos2unix & crypto++ dependencies
- added  gh_action_flatpak.sh flatpak builder script
- added com.solar2d.simulator.yml manifest for flatpak builder


airtonix:
Where are these flatpaks available for download? I can see in the daily build that the flatpak job runs and uploads an artifact...

PR #373, opened 2022-04-05
Batching fix: tightened up some "dirty state" rules

ggcrunchy:
The renderer tries to batch as much geometry as possible, to send it to the GPU all at once. One way to break such a batch is for certain states to be "dirty", e.g. a shader or texture has changed.

All the dirty state checks generally take the form (in Lua pseudo-code) `valueIsDirty = currentValue ~= newValue`. If _any_ value is dirty, the batch is broken.

Dirty state is handled and updated later on. In a few cases, this goes like so:

```
if valueIsDirty and newValue ~= nil then
  -- do something with the new value
  currentValue = newValue
end
```

In each situation, it's harmless to do nothing with the `nil` value, so not updating _is_ appropriate. However, at this point the batch has already been broken, even if only this ignored value changed. More seriously, since its current value stays the same, **EVERY** subsequent object that also a **nil** value will register as dirty.

To address this, the relevant dirty checks are rewritten as `valueIsDirty = currentValue ~= newValue and newValue ~= nil`.


ggcrunchy:
Also, forgot to describe a test.

You may change [this value](https://github.com/coronalabs/corona/blob/c1fdbb7f994c53d4f22278bfc0f5841f442bb75f/librtt/Renderer/Rtt_GLCommandBuffer.cpp#L31) from `0` to `1`.

You will then get a pretty noisy bunch of output.

What I do is build and run the simulator, then load a project. At that point, put a breakpoint [here](https://github.com/coronalabs/corona/blob/c1fdbb7f994c53d4f22278bfc0f5841f442bb75f/librtt/Renderer/Rtt_GLCommandBuffer.cpp#L668) and clear the console log once you get to it: some of that is from the welcome screen and the rest is just distracting.

Remove that breakpoint and put another [here](https://github.com/coronalabs/corona/blob/c1fdbb7f994c53d4f22278bfc0f5841f442bb75f/librtt/Renderer/Rtt_GLCommandBuffer.cpp#L959) and run to that.

This will show you what happened this frame.

In the case I used to diagnose this, I had an object whose shader used two userdata uniforms, followed by four colored rounded rects. These last four were all separate calls. When I stepped through it, it was getting stuck on those uniforms, which were `NULL` but not being updated accordingly.

I think something like this should reproduce the issue:

```
local object = display.newRect(100, 100, 50, 50)

object.fill.effect = "filter.colorMatrix"

-- the following both use uniform userdata: 
object.fill.effect.coefficients =
{
    1, 0, 0, 0,  --red coefficients
    0, 1, 0, 0,  --green coefficients
    0, 0, 1, 0,  --blue coefficients
    0, 0, 0, 1   --alpha coefficients
}
object.fill.effect.bias = { 0.6, 0.1, 0, 0 }

for i = 1, 4 do
  display.newRect(50 + i * 100, 200, 50, 50)
end
```

This _should_ result in two batches&mdash;these will show up as `Draw` messages in the log&mdash;but without the fix will have five.

PR #393, opened 2022-05-08
Remove unnecessary use of global variables

lfabreges:
Two global variables lineno and errorMsg were created in json_decode_file


Shchvova:
Great find! Thanks.

PR #400, opened 2022-05-29
Allow developers to use complex settings files

XeduR:
Attempting to load a more complex config.lua will raise a warning and attempting to do the same with build.settings seems to result in the file not being loaded entirely.

This "complex file" may be attributed to loading a single configuration file via `require()` in either file, so allow the developers to explicitly set their settings as a `complexFile` to allow them to be validated as normal.


platform/resources/ValidateSettings.lua
```diff
@@ -52,7 +52,8 @@ function validateSettings(arg1, arg2)
 
 		print = origPrint
 
-		if ok then
+		-- allow developers to optionally use more complex settings files without raising warnings
+		if ok or validSettings.complexFile then
```
Shchvova, line 56: `validSettings.complexFile` would always be true though, wouldn't it?

XeduR: You are absolutely right. Seems I had run my final tests in a debug build of the simulator, so I didn't catch the issue.

XeduR:
Currently in the PR:

- fixed one minor typo in 91615ba
- added support for including text tag "isComplexFile" anywhere in **config.lua**, which will allow Solar2D to skip the initial file validation and suppress outputting a warning regarding complex files. By the time the file is actually loaded and not just validated, then requiring external modules (and other functions) will work correctly. NB! This only works with config.lua and not with build.settings. 

Example usage in config.lua:
```lua
-- isComplexFile: suppresses initial file validation warnings.
local gameConfig = require("configs.gameConfig")

application =
{
	content =
	{
		width = gameConfig.contentWidth,
		height = gameConfig.contentHeight,
		scale = "letterbox",
		fps = gameConfig.fps,
	},

	license =
	{
		google =
		{
			key = gameConfig.licenseGoogle,
		},
	},
}
```

Since we're using `string.find()` to find the tag, it can be anywhere in the file as a simple comment.

XeduR:
In case someone finds this later, you can suppress the warnings in config.lua by just checking for any Solar2D specific Lua API in the global table.

As config.lua is loaded twice, first to conduct basic static file analysis, and a second time to read the file for the application's configuration, certain libraries like `_G.native` will not exist during the static file analysis phase. In other words, you can suppress the warnings by first checking `if _G.native then` and running your code afterwards.

PR #401, opened 2022-06-05
iOS: Added Keyboard Events

scottrules44:


platform/iphone/Rtt_IPhoneDevice.mm
```diff
@@ -393,7 +393,9 @@
 		case MPlatformDevice::kLocationEvent:
 		case MPlatformDevice::kHeadingEvent:
 		case MPlatformDevice::kMultitouchEvent:
-		case MPlatformDevice::kKeyEvent:
+        case MPlatformDevice::kKeyEvent:
+            hasEventSource = true; //There is not reliable way to check this
```
Shchvova, line 397: It seems this was added [around](https://developer.apple.com/documentation/uikit/uikeyboardhidusage/uikeyboardhidusagekeyboarda?language=objc) iOS13.3, so may be some check like
 `&UIKeyboardHIDUsageKeyboardA != NULL` would work. Also, got to test on the older iOS device

Shchvova (commented):
Looks good. May be improve availability check. I'll try to charge my older device and see if it works.

platform/iphone/Rtt_IPhoneDevice.mm
scottrules44: Just added 
`if (@available(iOS 13.4, *)) { //Key Events only work on iOS 13.4 >=
                hasEventSource = true;
}`

I tried &UIKeyboardHIDUsageKeyboardA != NULL but getting error "Cannot take the address of an rvalue"

PR #406, opened 2022-06-15
Android: Fix App Pinning Bug

scottrules44:


Shchvova:
Nice! How did you track it? I'll test and merge tomorrow

scottrules44:
Figured the issue has something to do with suspend state. When app was pinned resume state is not be triggered after closing and opening the app. Debugged until I found the System Monitor was using a deprecated api

PR #412, opened 2022-06-21
Transparent window support for Mac

ggcrunchy:
This offers preliminary support for transparent windows on Mac, roughly like what you see in the GIF [here](https://joeallenpro.com/support/transparent-window-mode/).

(Details were worked out from [this](https://stackoverflow.com/questions/16135485/can-the-background-for-a-mac-app-be-transparent), [this](https://forum.juce.com/t/simple-osx-trick-to-create-a-transparent-opengl-rendered-window/15506), [this](https://developer.apple.com/forums/thread/21626), and some guesswork after that.)

Basically, in your `config.lua`, you have something like

```
application = { isTransparent = true }
```

and somewhere in your code, do something like this, more or less my test:

```
-- Make transparent / translucent.
local alpha = 0 -- or .25, .5, etc.

display.setDefault("background", 0, 0, 0, alpha)

-- Some Solar objects above the background.
local random = math.random

for i = 1, 100 do
  local circle = display.newCircle(random(100, display.contentWidth - 100), random(100, display.contentHeight - 100), random(5, 15)

  circle:setFillColor(random(), random(), random())
end
```

and any pixels you don't draw will expose the background underneath the simulator / application (alpha = 0), or blend with it (> 0 and < 1).

As that last comment implies, I added support in the simulator to visualize the result, even though none of the devices themselves are supported. The Mac build, of course, will be appropriately transparent.

This is probably "experimental" at this point. Transparency requires a borderless window, which probably means some positioning / movement support is also needed to supplement it. (Maybe something akin to what the simulator skins do.)

(On that last note, a sanity check: do some of the skins just not show when running the simulator from a build? I was worried my changes were somehow causing this, but another branch seemed to show the same result.)


ggcrunchy:
This might also be interesting to explore: [NSVisualEffectView](https://developer.apple.com/documentation/appkit/nsvisualeffectview)

Issue #402, opened 2022-06-09
WinTimer crash on close / relaunch


ggcrunchy:
I believe this is quite an old issue, but I only isolated it just now from a Windows source build. It seems like if I leave a program running for a while (I don't know why this should matter, but it seems to) then attempt to relaunch / close Solar, it will often shut down soon afterward.

These always seem to be on <https://github.com/coronalabs/corona/blob/master/platform/windows/Corona.Native.Library.Win32/Rtt/Rtt_WinTimer.cpp#L78>, suggesting that the timer object has been destroyed and the memory there is bogus.

I confirmed that `Stop()` was being called, so this seemed strange.

The `KillTimer` docs (https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-killtimer) mention for the second parameter that it "must be the same as the `nIDEvent` value passed to `SetTimer`." so I thought <https://github.com/coronalabs/corona/blob/master/platform/windows/Corona.Native.Library.Win32/Rtt/Rtt_WinTimer.cpp#L61> just needed to be changed to `::KillTimer(fWindowHandle, (UINT_PTR)this);`.

The issue still occurs even with that "correction", though, so I'm a bit stumped. 😄

ggcrunchy:
Curiously, the timer class is mostly replicated over [in the network library](https://github.com/coronalabs/submodule-plugins-network/blob/696b05ae37c82bba6b6a6ee9a4639a77fb5250c8/win32/WinTimer.cpp). Among the differences is an ID -> timer map, whose entry is wiped in `Stop()` and then checked in the callback.

It would not surprise me at all if that was to account for this very problem; I'll give this a try.

ggcrunchy:
I've since submitted a [fix](https://github.com/coronalabs/corona/pull/414).

Issue #186, opened 2020-10-28
[Windows simulator] frequently crashes when relaunching simulator


XeduR:
This has been going on for quite a while already, long before Corona transitioned to Solar2D.

When relaunching the simulator via `Ctrl+R`, the simulator frequently hangs for several seconds and then it crashes. The simulator console remains open and shows the following message: `The stdin connection has closed.`

I haven't been able to find any particular reason as to what triggers this. Sometimes I get a day or two by without a single incident, whereas other days the simulator seems to crash between ~5-10 relaunches.

There seem to be two different types of the error messages in Windows Event Viewer, which are:

```
Source: Application Error
Task Category: (100)
Event ID: 1000
Level: Error

Faulting application name: Corona Simulator.exe, version: 20.0.3619.0, time stamp: 0x5f6dd8b5
Faulting module name: ucrtbase.dll, version: 10.0.18362.1110, time stamp: 0x100b54ae
Exception code: 0xc0000409
Fault offset: 0x0009efbb
Faulting process ID: 0x3600
Faulting application start time: 0x01d6ac9bb559023c
Faulting application path: C:\Program Files (x86)\Solar2D\Corona Simulator.exe
Faulting module path: C:\Windows\System32\ucrtbase.dll
Report ID: f1a7cd7b-ec5a-492a-a666-0473356e0446
```

```
Source: Application Error
Task Category: (100)
Event ID: 1000
Level: Error

Faulting application name: Corona Simulator.exe, version: 20.0.3619.0, time stamp: 0x5f6dd8b5
Faulting module name: VCRUNTIME140.dll, version: 14.24.28127.4, time stamp: 0x5d8e68be
Exception code: 0xc000041d
Fault offset: 0x0000317e
Faulting process ID: 0x1a1c
Faulting application start time: 0x01d6a0c2c18bc38f
Faulting application path: C:\Program Files (x86)\Solar2D\Corona Simulator.exe
Faulting module path: C:\Windows\SYSTEM32\VCRUNTIME140.dll
Report ID: 75276772-80f8-44da-a270-d91e1b0ab2b3
```

```
Source: SideBySide
Task Category: None
Event ID: 78
Level: Error

[...]

PR #414, opened 2022-06-23
Windows frame timer fix

ggcrunchy:
This is a fix for the issue brought up [here](https://github.com/coronalabs/corona/issues/402). Basically, the simulator will occasionally crash when you relaunch or close.

There's a possibility it's related to [this](https://github.com/coronalabs/corona/issues/186), but that will need some follow-up to confirm. I was testing using a project that did have audio, although I thought it had occurred in quiet ones too. In any case, it wouldn't be unreasonable that audio updates might exacerbate timer issues.

---------------------

**Details**:

Basically, it seems `KillTimer` does not actually evict relevant `WM_TIMER` messages from the queue, in which case they will still fire. There are vague hints of this in some online forum questions, but it's never spelled out. The problem in our case is that this might be after `WinTimer`'s destructor has gone off, so everything in our timer callback is garbage.

As mentioned in #402, it looks like the `network` library basically reused the `Rtt_WinTimer` code, apart from not assigning the timer to a window. Furthermore, it uses an ID &rarr; timer map, [using this in the callback](https://github.com/coronalabs/submodule-plugins-network/blob/696b05ae37c82bba6b6a6ee9a4639a77fb5250c8/win32/WinTimer.cpp#L26) to check if the timer still exists. I assume the same issue was being addressed.

I also mentioned in #402 that the wrong value was being passed to `KillTimer()`. This is per the [SetTimer docs](https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-settimer): "If the function succeeds and the _hWnd_ parameter is not **NULL**, then the return value is a nonzero integer. An application can pass the value of the _nIDEvent_ parameter to the `KillTimer` function to destroy the timer." and also those for [KillTimer](https://docs.microsoft.com/en-us/windows/desktop/api/winuser/nf-winuser-killtimer): "The timer to be destroyed. If the window handle passed to `SetTimer` is valid, this parameter must be the same as the _nIDEvent_
 value passed to `SetTimer`."

I don't know what happens if this is wrong, but I figure it didn't amount to much because the timer was basically used for the entire run of a Solar program and went away when the window did, on close / relaunch. But in theory `WinTimer` is now more robust for wider use, with the revisions in this PR.

The changes are a lot like what `network` does. A map from ID &rarr; timer makes sure spurious callbacks do nothing. I used an incrementing ID (starting from 1) rather than what `SetTimer` returns ("If the function succeeds and the _hWnd_ parameter is not **NULL**, then the return value is a nonzero integer."), since I'm not sure if that's recycled; for the same reason I didn't use a `std::set`, for the rare-but-possible case that another timer were reallocated into that same memory. These concerns are a bit academic with the timer being a singleton used from launch to close / relaunch, but it's not at all impossible that timers could be put to wider use. 😄

### DIFF
--- a/platform/windows/Corona.Native.Library.Win32/Rtt/Rtt_WinTimer.cpp
+++ b/platform/windows/Corona.Native.Library.Win32/Rtt/Rtt_WinTimer.cpp
@@ -15,7 +15,7 @@
 namespace Rtt
 {
 
-std::map<UINT_PTR, Rtt::WinTimer *> WinTimer::sTimerMap;
+std::unordered_map<UINT_PTR, Rtt::WinTimer *> WinTimer::sTimerMap;
 UINT_PTR WinTimer::sMostRecentTimerID;
 
 #pragma region Constructors/Destructors
--- a/platform/windows/Corona.Native.Library.Win32/Rtt/Rtt_WinTimer.cpp
+++ b/platform/windows/Corona.Native.Library.Win32/Rtt/Rtt_WinTimer.cpp
@@ -15,6 +15,9 @@
 namespace Rtt
 {
 
+std::map<UINT_PTR, Rtt::WinTimer *> WinTimer::sTimerMap;
+UINT_PTR WinTimer::sMostRecentTimerID;
+
 #pragma region Constructors/Destructors
 WinTimer::WinTimer(MCallback& callback, HWND windowHandle)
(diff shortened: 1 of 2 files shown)


Shchvova:
Moved to unordered_map just to prevent all the warnings.

PR #415, opened 2022-06-27
Upload flatpak release

### LINKED ISSUE #365: Publish to Flathub (follow-up from previous flatpak issue)
Hey! I'm very thankful for addressing the issue before regarding flatpaks as I find them to be wonderful. I suppose the last thing I'd want for users of them would be to upload it to the largest repository of flatpaks which is found on https://flathub.org/ both out of convenience and discoverability of the app as flatpak users would be able to find it in there. Alternatively, if you'd prefer setting up your own flatpak repo, that would be valid as well as flatpak's source is open for both client-wise and server-wise implementations so you could do either if you wanted. 

Cheers!
MetalMaxMX.

airtonix:
Includes the flatpak artifacts in the release

fixes #365

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -584,6 +584,7 @@ jobs:
       - Windows-Simulator
       - macOS-Simulator
       - linux
+      - linux-flatpak
       - CoronaCards-Android
       - CoronaCards-iOS
       - native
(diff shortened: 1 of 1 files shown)


Shchvova:
Hello! Unfortunatelly, flatpack currently fails to build, this is why I excluded it from the release (it was failing it).
You can check the latest build here https://github.com/coronalabs/corona/actions/runs/2536926166
As you see, flatpack artifact is 0 bytes.
And here why it fails to build: https://github.com/coronalabs/corona/runs/6990682787?check_suite_focus=true#step:10:236

Unfortunately, It builds locally on my VM and I already tried fixing it, but busy with other things right now. If you are failiar with flatpack quirks, feel free to help me out building it.

airtonix:
I was all enthusiatic up until i realised you've hardcoded everything so it only works on ubuntu.


i'm on fedora, so yeah... I can't be bothered setting everything up manually. thanks but i'll find another framework to use.

Issue #310, opened 2021-11-26
Mac Simulator: Crashes and closes when error with image file path or name/extension


jajern:
**Describe the bug**
If there is an error with a file name or path in the written code for display.newImage the simulator crashes. When it crashes it also closes the console so the only way I was able to find the bug was by running s screen recording of the console so I could replay the video and pause it when the error message displayed before the sim crashed.

**To Reproduce**
1. create an object for display.newImage
2. write in the wrong image file path or file extension
3. Simulator crashes

**Target platform and where build was made:**
 - Device: MacBook Pro 
 - OS: macOS 12.0.1

jajern:
I noticed an additional type of error that also crashed the simulator on Mac:

example:
newGroupName = display.newGroup()
group:insert( newGroupNam )

The error was I had the "newGroupName" misspelled inside the insert method and that threw the error and crashed the simulator. The only way to know what the error was, I had to do a screen recording of the console so I could see what the error was that flashed on the console before it crashed and closed.

Shchvova:
I couldn't reproduce this issue. Like, just having 
```lua
newGroupName = display.newGroup()
group:insert( newGroupNam )
```
Doesn't crash anything, I am getting regular error message saying 
```
ERROR: Runtime error
                    main.lua:2: attempt to index global 'group' (a nil value)
                    stack traceback:
                    	main.lua:2: in main chunk
```

jajern:
Thank you for testing on your end. I start noticing these crashes on a new MacBook with the m1 chip. I recently transferred from an Intel MacBook and did not have these problems on the Intel.

I am thinking these crashes could either be related to the simulator running on Apple Silicon MacBook, or some issue with transferring files/preferences from one Mac to another. 

Is the machine you are using a MacBook with an M1 chip?

scottrules44:
This should be fixed new build

jajern:
Hi guys, I saw this was marked as fixed in build 3675. I am currently using build 3677 and am still experiencing Simulator crashing with errors. It is not happening with all errors and it is happening with different kinds of errors.

For example one type of error that caused the crash was when inserting an object into a group, the object name was misspelled in the insert code: 
groupName:insert(misspelledObjectName)

Another type of error that caused the crash was code inside of the onRowRender function for a widget newTableView. The code inside the renderer was creating a newText object and the reference to the table was incorrect:
local rowTitle = display.newText( row, fullItemsList[row.index][1], 0, 0, nil, 14 )
[...]

PR #422, opened 2022-07-08
MacOS: Fixes Simulator Crashes on Error for M1(ARM)

scottrules44:


scottrules44:
Should fix issue: #310 (I think I may have seen similar issues)

Shchvova:
OK. Lets give it a try

PR #434, opened 2022-08-02
Android: adding settings.intentFilters.autoVerify

Shchvova:


solares:
@Shchvova Thanks. We'll test and report back.

PR #438, opened 2022-08-17
New display default for external textures (minor texture issues with certain effects)

ggcrunchy:
A couple people were testing a plugin of mine and noted that it would produce different results with the pixelate filter.

I looked into it, and will quote from my response to one of them in summary:

"Okay... external textures (Bytemap, etc.) seem to be classified as retina. This affects the content scaling uniform.
https://github.com/coronalabs/corona/blob/master/librtt/Display/Rtt_TextureFactory.cpp#L577 (the `true`).
So, it goes through this path https://github.com/coronalabs/corona/blob/master/librtt/Renderer/Rtt_Renderer.cpp#L625 versus the other, and indeed that's the relevant part: https://github.com/coronalabs/corona/blob/master/librtt/Display/Shader/kernel_filter_pixelate_gl.lua#L31"

This PR adds a new value for `display.setDefault()`, **"isExternalTextureRetina"**, that may be set to alter the default behavior.

This is a very slight modification of the original test, just switching out a quick-and-dirty image for the original assets:

[pixelateBytemapTest.zip](https://github.com/coronalabs/corona/files/9366507/pixelateBytemapTest.2.zip)

As is, it will demonstrate the problem. Uncommenting the line that sets the default to `false` will fix it.


ggcrunchy:
Uff, I've botched this royally, trying to bring it back in sync. 😃Will start afresh, now that I have a better base.

PR #453, opened 2022-09-26
Simulator: Fix copy to device for Android

scottrules44:
Using "Copy to device and launch" (Android) does not work with aab


Shchvova:
Ah...  my bad

Shchvova (approved):

PR #473, opened 2022-11-04
Linear Guassian blur effect

ggcrunchy:
@XeduR and I were discussing the current Gaussian blur effect, which can be pretty expensive, especially on many mobile platforms. I implemented the idea from [this RasterGrid post](https://www.rastergrid.com/blog/2010/09/efficient-gaussian-blur-with-linear-sampling) and it seems to work quite well, albeit with slightly different characteristics.

The stock Gaussian effect is able to do quite a large loop and must be able to branch out of that as well, probably explaining the cost in large part. This variant, on the other hand, will do five samples on each directional pass, all straight-line code.

The new effect is available as "filter.blurGaussianLinear".

I'm not sure how useful it is, but I made the offsets available as parameters `offset1` and `offset2`, defaulting to those listed at the end of the article.

[Here's](https://github.com/coronalabs/corona/files/9940699/TransitionLagIssue.zip) the test project.

I'm having a little difficulty at the moment working out the incantations to launch iOS and tvOS test builds, but confirmed it works on Windows, Mac, and Android. I've not yet built anything on web or Linux, but also made an attempt to put the new files in place in their respective projects, from eyeballing what other effects are doing.


Shchvova (approved):
Wow. Nice and clear. 
kudos for clean 🧼 PR

Shchvova:
TODO:

- [ ] add docs: https://github.com/coronalabs/corona-docs/blob/master/markdown/guide/graphics/effects.markdown
- [ ] add to samples: https://github.com/coronalabs/samples-coronasdk

PR #475, opened 2022-11-05
 fixing null tracking rect tag

Shchvova:


Shchvova:
this bug never made it to the build and shoud've been part of #474

PR #484, opened 2022-12-03
Android: fixing vibration style being ignored

Shchvova:


scottrules44:
My mistake; the suggested change fixes the issue

appfamily:
Thanks @scottrules44 , does that mean all the different vibration types in the test app work?

Shchvova:
I think it would be easiest to merge this and see if it fixes your issue, if not, we'll make another build

Issue #260, opened 2021-07-02
Feature request: Implement "Include Standard Resources" build option for all platforms


XeduR:
With the HTML5 builds, there's already an option to include (or to not include) standard resources when building using the Solar2D Simulator.

![include standard resources](https://user-images.githubusercontent.com/42606839/124257247-dc8de080-db34-11eb-9b03-04d58f44b04b.png)

If I'm not mistaken, including the standard resources consists of loading most of the files from here: https://github.com/coronalabs/framework-widget (hopefully not including the unit tests, psd files, etc.)

In any case, the images alone are around 800-900KB, and with the Lua files, these amount to around 1.5MB of files that are never used by most games/apps.

Would it be possible to make this "Include Standard Resources" build option available for all platforms, not just HTML5?

PR #486, opened 2022-12-12
macSim/iOS: Added option to remove std resources

scottrules44:


Shchvova (approved):

sekodev:
Is this related to #260 ?

XeduR:
Indeed it is, this is a long-requested feature for the Windows simulator as well, though on Windows you can also get rid of the files after building by using batch scripts and such.

Shchvova (approved):

solares:
Are we able to set this field using CoronaBuilder to build? We are always looking at ways of reducing app size.

Issue #301, opened 2021-11-15
print(os.date("%s")) crashes the sim


mark-sgs:
Place `print(os.date("%s")) crashes the sim` in main.lua and it crashes.  It should raise an error at the very least,

XeduR:
This quite a strange crash indeed.

It also doesn't seem to crash a built project. For instance, you can run the same code over at https://www.solar2dplayground.com/ and it'll just print "%s".

scottrules44:
Is this only happening on Windows? I just tried on my Mac, cannot reproduce

XeduR:
Yeah, we're both using Solar2D Simulator on Windows.

scottrules44:
This is fixed with 2022.3685+

PR #490, opened 2022-12-20
Windows: fixes os.date crashing

scottrules44:
This address #301

Screenshot 2022-12-20 at 2 17 32 PM


Shchvova (approved):
seems to work for other folkx

PR #495, opened 2023-01-05
MacOS: Prevent Lua Crashes on Apple Silicon

scottrules44:
Was working on a project with the newest version of Solar2D and got a crash on my Mac. Was able to prevent the crash by checking if my device is Apple Silicon(arm) and preventing `raise` from being executed.


Shchvova:
We would probably want to. Ache result of the function would get reused. This will work for now.

Shchvova (approved):

PR #523, opened 2023-02-28
Unlink libjpeg from Android builds

### LINKED ISSUE #522: [Android] The Android platform does not seem to need to compile libjpeg?
Because some Android stores require to eliminate CVE as much as possible, so I paid attention to libjpeg, a three-party library
And, the result of display.newImage("1.jpg") is correct.
The implementation of display.save(xx,"xxx.jpg") seems to be independent of libjpeg, but I haven't tested it.

is it possible to remove it from the CMakeList file. 
image

Removing libjpeg can reduce the so file size by about 4% 
better than nothing  :)
image

CVE in libjpeg
image

scottrules44:
Based on Issue #522, (Fixes #522)
Android does not seem to rely on libjpeg from my research, only seems to use BitmapUtils which are primarily used for Linux and Emscripten(html)


zhiyangyou:
It may be necessary to modify this file?
**/platform/android/ndk/CMakeLists.txt**

Shchvova:
We’re using CMake for Android. Mk files can be deleted afaik.
Here is the main file https://github.com/coronalabs/corona/blob/master/platform/android/sdk/CMakeLists.txt


And here are where supplementary file with source file list has reference to libjpeg
And a
https://github.com/coronalabs/corona/blob/648b1dde2e3bbd90197cc37c8ede27b1db498485/platform/android/ndk/CMakeLists.txt#L728

Shchvova:
I pushed commit with required changes. It may I ask you to test it?

Shchvova (approved):

Shchvova:
I decided to delete bunch of files... Seems appropriate, they are of no consequence.

zhiyangyou:
my testcase: 

- build (✔) use **platform/android** project

- display.newImage(gXXX,"1.jpg") (✔)

- display.save(gXXX,"1.jpg") (✔)

zhiyangyou:
> I decided to delete bunch of files... Seems appropriate, they are of no consequence.
these?
xxxx.vcxproj 
xxxx.vcxproj.filters
in **platform/android/sdk/...**

Shchvova:
Oh... one of our developers really into Visual Studio. So just ignore those.

Issue #269, opened 2021-08-10
[Core]: Touch event xDelta and yDelta are incorrect


XeduR:
xDelta and yDelta properties were added to touch event in https://github.com/coronalabs/corona/commit/68e90cbdafb50c07be5b850c13fc02f8d75e5c9c, however, they were implemented incorrectly.

The documentation pages for [xDelta](https://docs.coronalabs.com/api/event/touch/xDelta.html) and [yDelta](https://docs.coronalabs.com/api/event/touch/yDelta.html) have example codes that only work if the the device/window dimensions are identical to those of config.lua.

**The reason is that these delta values are both scaled by Solar2D and then rounded to the nearest integer (neither of which should be happening).**

For example, if you have a config.lua content area of 480x320 and you run the simulator on a 960x640 device, then the [scale factor](https://docs.coronalabs.com/guide/basics/configSettings/index.html#dynamic-image-selection) will be 2. This means that when you drag a touch event by 10 pixels, this gets multiplied by 2, to 20. Or if you simulate a 1920x1080 device, then the scale factor will be 1.6875, so the 10 pixels gets scaled to 16.875, which then gets rounded to 17.

The real problem occurs when Solar2D also scales the display object movements, so if you have a 480x320 content area and you drag a display object by 10 pixels on a 960x640 display, then the delta values get scaled to 20 and then the display object movement gets scaled again to 40, resulting in double scaling/movement.

Currently, you can obtain the correct delta values by:
```lua
xDeltaReal = event.x - event.xStart
yDeltaReal = event.y - event.yStart
```

In summary:
- `event.xDelta` and `event.yDelta` values are being rounded, which should not happen.
- `event.xDelta` and `event.yDelta` values are being dynamically scaled, which should not happen.

PR #525, opened 2023-02-28
Core: Touch event xDelta and yDelta Fixes

scottrules44:
Addresses #269

### DIFF
--- a/librtt/Rtt_Event.h
+++ b/librtt/Rtt_Event.h
@@ -863,6 +863,7 @@ class HitEvent : public VirtualEvent
 
 	protected:
 		static void ScreenToContent( const Display& display,  Real xScreen, Real yScreen, Real& outXContent, Real& outYContent );
+        static void ContentToDelta( const Display& display, Real xContent, Real yContent, Real& outXDelta, Real& outYDelta );
 		bool DispatchEvent( lua_State *L, HitTestObject& parent ) const;
 
 	private:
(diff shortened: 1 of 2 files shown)


Shchvova:
I am worried about this one... It would break existing games, wouldn't it?

scottrules44:
Well it has to due to that event.xDelta and event.yDelta were implement wrong it is most likely they were ignored.
It is important to note the **relatively** new addition (xDelta and yDelta were added back in 2019)
It is possible that handful of people used the delta value and worked around the wrong values. 
Either was the values are wrong. I would argue that this fix does way more good than harm

scottrules44:
Also I scanned all 44 cases publicly available games on Github and non of them should be affected by the change. It will mainly affect games that have implement a crazy workaround get xDelta/yDelta

Shchvova:
Ok. I'll test an merge then

scottrules44:
^Figured out there is a much simpler way to implement, also here attached a demo project you can test with, should compare the values
[assets2.zip](https://github.com/coronalabs/corona/files/10867866/assets2.zip)


```
print("Correct Delta x and y values:")
local dx = event.x - event.xStart
local dy = event.y - event.yStart
print( dx )
print( dy )
print("Delta x and y values:")
print(event.xDelta)
print(event.yDelta)
```

librtt/Rtt_Event.h
```diff
@@ -945,8 +945,8 @@ class TouchEvent : public HitEvent
 		mutable Real fXStartContent;
 		mutable Real fYStartContent;
 		Real fPressure;
-		Real fDeltaX;
-		Real fDeltaY;
+        mutable Real fDeltaX;
```
Shchvova, line 948: Is this still required?

scottrules44: It is not, just removed :)

scottrules44:
Any updates on this pull @Shchvova :) ?

clang-clang-clang:
Any updates on this pull? Thanks.

PR #529, opened 2023-03-03
Windows: After Build options for Android

scottrules44:
Took a bit of work since I am not really a Windows Guy :)
Screenshot 2023-03-03 at 11 06 05 AM


platform/windows/Corona.Simulator/BuildAndroidDlg.cpp
```diff
@@ -812,14 +815,30 @@ void CBuildAndroidDlg::OnOK()  // OnBuild()
 		return;
 	}
 
+	//After Build Options
+	if (afterBuild == AB_COPY_TO_DEVICE) // Copy to Device
+	{
+		if (PathFileExists(_T("C:\\Program Files (x86)\\Android\\android-sdk\\platform-tools\\adb.exe"))) { // this should be where adb is installed
```
Shchvova, line 821: we actually install android sdk when building for android on Windows
`%APPDATA%\Corona Labs\Corona Simulator\Android Build\sdk\platform-tools\adb.exe`
Also, make sure APK exists. When AAB is generated, APK won't be created.

scottrules44: Gotcha will check this

scottrules44: I just pushed a fix for this and added apk check to make sure it exists.

I also removed the adb file check since it is bundled with Sim.

PR #531, opened 2023-03-05
MacOS Sim: Allow for dragging main.lua file on Sim

scottrules44:
Inspired by Window Version, Mac Sim could open a project by dragging main.lua inside window

drag


scottrules44:
Two quick notes, I also updated some old copyright dates while I was messing around, and this only works on the welcome screen (not in a skin) which matches windows and avoids confusions

Shchvova:
Lol. Should've had maeged this PR before all this got pushed in :D what's up with that?

PR #560, opened 2023-04-30
Maintenance

scottrules44:
Bump iOS/tvOS to 11.0 and Remove Xcode_13.2.1 since Apple is removing support for this "Starting April 25, 2023, iOS, iPadOS, and watchOS apps submitted to the App Store must be built with Xcode 14.1 or later. The latest version of Xcode 14, which includes the latest SDKs for iOS 16, iPadOS 16" 
https://developer.apple.com/news/?id=jd9wcyov


Shchvova:
I implemented configurable targets in #561, as well as building native on Xcode14.3, which also means building macOS template on it.

PR #563, opened 2023-05-01
Remove Xcode 13

scottrules44:
Remove Xcode 13.2.1 since Apple is removing support for this "Starting April 25, 2023, iOS, iPadOS, and watchOS apps submitted to the App Store must be built with Xcode 14.1 or later. The latest version of Xcode 14, which includes the latest SDKs for iOS 16, iPadOS 16"
https://developer.apple.com/news/?id=jd9wcyov

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -153,54 +153,8 @@ jobs:
           path: ./output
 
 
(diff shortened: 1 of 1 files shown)


Shchvova:
we don't need to do this yet

Issue #533, opened 2023-03-13
[Win32]: display.newText() , text rendering errors in RTL languages


zhiyangyou:
As shown in the screenshot, there is a piece of Arabic text, and it can be clearly observed that the position where the number '3' appears is inaccurate.

Then I modified the logic here to determine whether the string contains RTL characters. If so, add the DT_RTLREADING flag to the flags variable and adjust the alignment attribute logic. It seems to work normally.
Judging whether a string is mainly RTL text seems to be a very troublesome thing. There are some very complicated implementations in the ICU library. Of course, I only did a simple process here, that is, as long as it contains RTL unicode characters, It is considered to be RTL text, but this does not seem to be a perfect approach.

**Test text:**
**مرحبًا ، اسمي # name # ، عمري 3 سنوات هذا العام ، وأتيت إلى هذه الروضة للذهاب إلى المدرسة ، آمل أن تعطيني المزيد من النصائح!**

**Where to modify the code:**
https://github.com/coronalabs/corona/blob/2c211d807a5ef039b6f885b2deb85f49b9a3f68e/platform/windows/Corona.Native.Library.Win32/Rtt/Rtt_WinBitmap.cpp#L670
**ScreenShot:**
![84a7095c99159e4996dbb5b70b9dffbc](https://user-images.githubusercontent.com/70576712/224646712-e8ad8915-5627-4ef9-93a2-37dd119f66e8.png)
**Code modify screen Shot:**
![image](https://user-images.githubusercontent.com/70576712/224649316-1cde5868-f738-4be1-bf48-4d12b61b5465.png)

**Judge RTL character code snippet**
`	
        // aft_isrtl impl:
	//1  https://stackoverflow.com/questions/5423960/how-can-i-recognize-rtl-strings-in-c
	//2  https://www.appsloveworld.com/cplus/100/247/how-can-i-recognize-rtl-strings-in-c  
 

	
	bool ContainsRTLCharacters(const WinString& winStr)
	{
		const wchar_t* str = winStr.GetUTF16();
		size_t len = wcslen(str);
		for (int i = 0; i < len; i++)
		{
			wchar_t c = str[i];
			if (aft_isrtl(c))
			{
				return true;
			}
		}
		return false;
	}`

scottrules44:
See https://github.com/coronalabs/corona/pull/579


But the suggestion this would not constant across platform

You can change text alignment already 
https://docs.coronalabs.com/api/library/display/newText.html#align-optional

PR #579, opened 2023-05-27
Win32: display.newText() , text rendering errors in RTL languages

scottrules44:
Suggestion based on #533


Shchvova:
Is this consistent with macOS?

scottrules44:
Good question, will double check, but this mainly for Windows

Shchvova:
It is complicated topic to decide this things for user to be honest.. btw, if whole text is ltr, doesn't it render it properly?

PR #592, opened 2023-06-28
Separation of some classes into their own files; removing some header dependencies; other

ggcrunchy:
The capture textures and vertex format extension lists were piggybacking on some other files: canvas textures in the former case, and geometry renderer (and friends) in the latter. I did this as a combination of not quite knowing what all needed to be done to add new files on all platforms (mostly for captures) and early design, in the latter.

In addition, the extension lists and related types are now `class` rather than `struct` types, with some redesign around the same.

---

Configured Vulkan to the `PreRelease` configuration.

---

There was a bug in `Array::PadToSize()`, where it wasn't copying over the old memory when it grew!

---

I'd attempted to rewrite `Rtt_NEW` in Debug to use the memory leak checking variant, but looks like it doesn't play well with GDI bitmaps, so it's hidden by a `#define` guard.

---

Extension lists now compare with a hash, so I added a method to generate one in `String`. (It has different usage than `StringHash` itself.)

I ended up adding a different hash in what will be the follow-up PR. In this case the use cases might be similar enough that I'll just pick one or the other. 😄

---

I don't know if this was actually a stated goal, but `"CoronaGraphics.h"` and such didn't seem to be included by the `Renderer` headers. I did a little redesign to follow that.

Similarly, I switched some `CoronaLog()` calls to use `Rtt_TRACE_SIM()` instead.

---

Removed "custom ops" and dummy bounds routine that I missed earlier.


Shchvova (approved):
looks good to me.

PR #593, opened 2023-07-22
Support for Android api 33

scottrules44:


Shchvova:
Oh cool. I am working on this too. You migrated to android build plugin v7. Does packaging still works for you?

scottrules44:
Yes was able to make a test app from Android Studio, I have not tested packaging into Corona Sim and building with Sim. But can give it a try :)

Shchvova:
Make sure to try including assets like images and sounds.

scottrules44:
Just tested with 'AudioPlayer' demo and does include graphics and music

scottrules44:
'NativeDisplayObjects' demo seems to work as well, I also did test back button as well and seems to work normally with Android 13

Shchvova:
I will go through test cases and merge this on the week. If you want to help with testing even more consider trying out push notifications.

scottrules44:
> I will go through test cases and merge this on the week. If you want to help with testing even more consider trying out push notifications.

Will do!

scottrules44:
The Local Notification Demo works (Tested with my custom Corona Sim)

We will have to tweak the plugin to request include the new permission
 
`timer.performWithDelay(3000, function()

  local options =
  {
      appPermission = "android.permission.POST_NOTIFICATIONS",
      urgency = "Critical",
      listener = appPermissionsListener,
      rationaleTitle = "Notification access required",
      rationaleDescription = "Notification access is required to send notifications",
      settingsRedirectTitle = "Alert",
      settingsRedirectDescription = "Without the ability to take notification, this app cannot properly function. Please grant notification access within Settings."
  }
  native.showPopup( "requestAppPermission", options )
end )
`

Screenshot_20230726-145639

Shchvova:
So... I did quite a bit of testing and I have good and bad news - good news that it kinda works. Bad news - back button stuff is broken. To test it you need to set ` <application android:enableOnBackInvokedCallback="true" />` as enable "Predictive back animations" in Developer settings on device. If you do that one can see that back functionality breaks in webPopUp, and generally whent trying to intercept system back button event. I think more thoughfull code should be added to do that. Luckuly we can opt out from it by setting that setting to 'false'

About Notifications, I think we should implement the popup if user tries to schedule a notification without a prior request, or tries to register for notifications. I will take care of it.

Great job, @scottrules44. Thanks a lot for taking a look into it.

Shchvova (approved):

PR #603, opened 2023-08-17
Added getSelection for native.textField

DimaUndSo:
I was working on a custom pixel art textfield for our game and needed a function that can return the current position of the cursor/selector of a native.textField.
There is a function "setSelection" with which you can set the cursor to a position, but sometimes you just need to know the current position/selection.
So I added the function getSelection below the setSelection function for ios, android, mac and windows.
I've tested all platforms.
I also added support for setSelection for macOs Simulator. There was a Warning before that it's not supported, now it is.

That's how you use it:
`local start, finish = myNativeTextField:getSelection()`
This will return you the cursors/selectors first and last position.

Hope you like it and can add this to a daily build soon, so I can normally use it in our game.
Thanks

### DIFF
--- a/platform/android/ndk/Rtt_AndroidTextFieldObject.cpp
+++ b/platform/android/ndk/Rtt_AndroidTextFieldObject.cpp
@@ -125,6 +125,26 @@ AndroidTextFieldObject::setSelection( lua_State *L )
 	return 0;
 }
 
+int
+AndroidTextFieldObject::getSelection( lua_State *L )
+{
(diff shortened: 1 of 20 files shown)


Shchvova (approved):
Nice job!

sekodev:
@DimaUndSo does this also work for TextBox? I see you've also made changes to text box code but you haven't mentioned it in your description.

DimaUndSo:
> @DimaUndSo does this also work for TextBox? I see you've also made changes to text box code but you haven't mentioned it in your description.

I replied you here: https://github.com/coronalabs/corona-docs/pull/152

PR #609, opened 2023-08-22
Android: keep resource as is

clang-clang-clang:
> The option setting 'android.enableResourceOptimizations=false' is deprecated. The current default is 'true'. It will be removed in version 8.0 of the Android Gradle plugin.

^ Although the Android Gradle plugin will complain, but afaik Solar2D need to keep the widget resource as-is before found out more practical solution.

### DIFF
--- a/platform/android/gradle.properties
+++ b/platform/android/gradle.properties
@@ -5,3 +5,4 @@ org.gradle.jvmargs=-Xmx1280m
 # Migrate To Android X
 android.useAndroidX=true
 android.enableJetifier=true
+android.enableResourceOptimizations=false


Shchvova (approved):

clang-clang-clang:
Refer: commit ad9bf162e2db195a5d8d4682336009b20db1d072 and commit f12fb5748dc1fe3f98610101be4647be4f3eabf8

Without the option, the default value of "true" is used, which results in the resource filenames being shrink and obfuscated, resulting in the widget resource not being found.

Set the option to `false` is temporary solution (this PR), may find another solution later (#610 as memo) :D

clang-clang-clang:
follow up on #611

Shchvova:
Haha :) I am the one who put it there in the first place.

Issue #634, opened 2023-10-11
[Apple]: runtime bug, network EXC_BAD_ACCESS


clang-clang-clang:
- [x] I tried searching [Solar2D forums](https://forums.solar2d.com/) for the solution, still open:
  - https://forums.solar2d.com/t/network-request-and-ios-17-app-crashes/356042
  - https://forums.solar2d.com/t/simulator-crash-on-mac-os-sonoma/356036

**Describe the bug**
In plugin/network/apple, `NetworkRequestState`'s `fRequestURL` EXC_BAD_ACCESS in `pushToLuaState`.

**To Reproduce**
- Run one of [network.request()](https://docs.coronalabs.com/api/library/network/request.html) samples on 3698 at macOS Sonoma;
- Or run with platform/iphone project's `template` built by Xcode 15 at iOS 17 device;
- Or run with platform/mac project's `rttplayer` built by Xcode 15 at macOS Sonoma.

**Possible Solution**

```diff
plugin/network/apple/AppleNetworkSupport.mm
 	fPhase = @"began";
 	fStatus = -1;
-	fRequestURL = url;
+	fRequestURL = [url copy];
 	fResponseHeaders = nil;
 	fResponseType = @"text";
```
iOS will take effect, but macOS need run `build.sh` first then copy network.dylib from build-core.

**About License**

Are the [submodule-plugins-network](https://github.com/coronalabs/submodule-plugins-network) and [submodule-plugins-build-core](https://github.com/coronalabs/submodule-plugins-build-core) and all the submodules follow up this main repo's MIT license? @Shchvova May I ask to setup a proper LICENSE on those repos?

clang-clang-clang:
Related to https://github.com/coronalabs/corona/issues/632.

clang-clang-clang:
Possibly relevant Xcode 15 Release Notes:
## [Apple Clang Compiler](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#Apple-Clang-Compiler)
### Resolved Issues
- Fixed: Improvements in the ARC optimizer allow Objective-C and Swift objects to be released earlier. ...

clang-clang-clang:
Fixed in [3699](https://github.com/coronalabs/corona/releases/tag/3699).

PR #636, opened 2023-10-13
Apple: Fixes for networking for iOS 17 and MacOS 14

scottrules44:


scottrules44:
Fix based on #634

Shchvova (approved):

Shchvova (commented):
I think this will leak memory. Let me tinker with

Shchvova:
In future please try to avoid making PRs from your `master`. It messes with repos (mostly yours)

Shchvova:
I accidentally merged this PR, actual fix is in 11bff30cb83b1d73b742f76c33340a98319176d1

clang-clang-clang:
It works.
Your solution is better, but strangely it didn't work when I tested `retian` only for `fRequestURL` :(, however I incorrectly assumed that ARC would automatically reclaim the `copy` object.
Test case passed on iOS 17 and macOS Sonoma built with Xcode 15. 
Thank you. close #634.

PR #648, opened 2023-11-14
Android maintenance

clang-clang-clang:
This PR contains where

# Fix typo
Some typo fixes.

# Fix Gradle task not recognize file changes
Fix when using the 'Run' button in Android Studio to re-run with a changed Lua project file, the Gradle task does not recognize the changes and **incorrectly** marks mergeDebugAssets as `UP-TO-DATE`, this cause App remain unchange.
This is a temporary solution to facilitate debugging of Android applications by Solar2D engine maintainers, and Gradle tasks should be reorganized in the future.

#  FileService IO buffer size changed
Adjust buffer sizes that are more friendly to memory paging and I/O throughput. 64KB size vs. the original 1KB, using FileService#getBytesFromFile() reading 100MB (unflatten) of `resource.car` is at least a 10% improvement on the Pixel 7 and 30% on the Galaxy S7 edge.
Additionally, in the future, it will be necessary to change the way Android reads Lua code packages (`resource.car`) to the same mmap method as Apple and Linux platforms, in order to save a copy of runtime memory of the size of `resource.car` after it is unflatten.

# Remove redundant splash resource check
In an ANR problem we encountered, we found that CoronaActivity#showCoronaSplashScreen() initializes the FileService to do the caching of the relative location of the file to improve the efficiency of the subsequent reading of the resource, this IO operation is performed in the main thread resulting in an ANR, and it is not necessary to retrieve the existence of the resource from the FileService while checking the resource ID, choosing one or the other is sufficient.
See [getIdentifier()](https://developer.android.com/reference/android/content/res/Resources#getIdentifier(java.lang.String,%20java.lang.String,%20java.lang.String)).

# loadExpansionFiles even when storage read-only
.

# Introduced reportFullyDrawn()
See [reportFullyDrawn()](https://developer.android.com/reference/android/app/Activity#reportFullyDrawn()) and [TTFD](https://developer.android.com/topic/performance/vitals/launch-time#retrieve-TTFD), users can call `system.request("reportFullyDrawn")` to report Fully Drawn.

Feel free to raise the discussion.
Thanks.

### DIFF
--- a/librtt/Rtt_Runtime.cpp
+++ b/librtt/Rtt_Runtime.cpp
@@ -1963,7 +1963,7 @@ Runtime::operator()()
 	const bool isSuspended = IsSuspended();
 	if( wasSuspended != isSuspended && isSuspended )
 	{
-		// This condition is writtein inverse for better undrerstanding
+		// This condition is written inverse for better understanding
 		// Sometimes (Splash Screen is shown) scheduled tasks can suspend Runtime
 		// In that case (suspension state is changed and it is suspended), skip Display update
 	}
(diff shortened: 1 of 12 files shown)


Shchvova (approved):
Looks good. I'll iterate on build file and merge later.

Shchvova (approved):

PR #655, opened 2023-12-01
Core/Android: add sanity check

clang-clang-clang:
> Does it make sense to add these checks in the case of OOM? Should we "fix" these boundary cases? Or should we just expose them to expose the game code error?
> 
> ^These questions bother me, but I'm willing to give it a try, so if the changes don't fit please feel free to discuss.

These fixes come from analyzing the OOM crash stack, enhancing some of the checks.
These changed stack contents in cases where an error had already occurred.

Regarding`Data.Set()`, it is envisioned that in the OOM case, fLength retains a non-zero value even if fStorage is empty. It is good practice to perform a `Data.Get()` check before using it.

Of course, there's also an Android `build.gradle.kts` issue where switching `coronaSrcDir` causes the previous project's `resource.car` content to be incorrectly retained as the gradle script uses the `CoronaBuilder car` command to do the _append_ action.

This time I modified the file header, replaced Corona with Solar2D and added declarations.

Thanks.

### DIFF
--- a/platform/android/app/build.gradle.kts
+++ b/platform/android/app/build.gradle.kts
@@ -487,6 +487,8 @@ android.applicationVariants.all {
             }
             delete(metadataConfig)
             val toArchive = outputsList.filter { file(it).name != "config.lu" } + "config.lu"
+            // Make sure it's not appended to an old file
+            delete(compiledLuaArchive)
             mkdir(file(compiledLuaArchive).parent)
             exec {
                 workingDir = file(compiledDir)
--- a/librtt/Core/Rtt_Data.h
+++ b/librtt/Core/Rtt_Data.h
@@ -1,6 +1,7 @@
 //////////////////////////////////////////////////////////////////////////////
 //
-// This file is part of the Corona game engine.
+// This file is part of the Solar2D game engine.
+// With contributions from Dianchu Technology
 // For overview and more information on licensing please refer to README.md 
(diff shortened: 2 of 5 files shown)


Shchvova (approved):

platform/android/app/build.gradle.kts
```diff
@@ -487,6 +487,8 @@ android.applicationVariants.all {
             }
             delete(metadataConfig)
             val toArchive = outputsList.filter { file(it).name != "config.lu" } + "config.lu"
+            // Make sure it's not appended to an old file
+            delete(compiledLuaArchive)
```
Shchvova, line 491: do we still need it after recent changes? https://github.com/coronalabs/corona/commit/82d836f881d632f7100b55f8246a471c21609600#diff-60d531036c673f341a8267c4d40e0545ae42344ce4385abcb7ef5b28cc92fb7b

clang-clang-clang: Yes. It's different.

Shchvova: I approved this PR. Feel free to merge.

Shchvova: Or should I?

clang-clang-clang: Yes. Of course.
hah, those are the changes I can made so far.
Thank you.

Issue #296, opened 2021-10-30
[PLATFORM]: runtime bug


nobstudio:
- [ ] I tried searching [Solar2D forums](https://forums.solar2d.com/) for the solution.
- [ ] https://forums.solar2d.com/t/audio-oncomplete-bug-caused-crash/354132/2

**Describe the bug**
Hi, I can confirm there is a bug with audio.play onComplete, it caused crashes on some android phones. Even if the function is empty, it will crash. It only happens to 10% of the players, I use same phone model, same android 11, couldn't reproduce the bug.
My game on GooglePlay had 800+ crashes in 6 days, after I removed onComplete for all audio, 0 crash.

**To Reproduce**
use a short 0.3 second sound, and let the simulator run for a while(30 sec to 10 minutes). On windows, the simulator will force quit with this error **The stdin connection has closed.**
```
mySound = audio.loadSound("mySound.wav")
        soundTesting = function()
end
function checkSound ()
	for j = 1, 30 do
		audio.play(mySound , { onComplete=soundTesting })
	end
end	
Runtime:addEventListener( "enterFrame", checkSound )
```

**Target platform and where build was made:**
 - Device: Android, windows simulator
 - OS: android 9,10,11

**Additional context**
The error from googleplay is 
```
signal 11 (SIGSEGV), code 2 (SEGV_ACCERR)
libcorona.so!libcorona.so (offset 0x37e000)
Rtt::ALmixerSoundCompletionEvent::~ALmixerSoundCompletionEvent()) (SourceCode: /Users/runner/work/corona/corona/librtt/Rtt_PlatformOpenALPlayer.cpp:76)
```

alirezapoorslt:
Why you need to play a sound 30 times in each frame?

nobstudio:
> Why you need to play a sound 30 times in each frame?

to speed up testing. 30 times each frame to stress test and reproduce this error in a short period of time.
For my game, only up to 10 audio channels are used at all times, and I can't get this error on my device or windows simulator at all. But for some players, they have this error every few minutes, not sure what's wrong with their devices.

sekodev:
Is it possible issue #19 and this one is related?

PR #657, opened 2023-12-05
Possible fix for audio.onComplete

ggcrunchy:
I took the sample in [this issue](https://github.com/coronalabs/corona/issues/296) and seemed to pin down where the crash itself was happening. A description and analysis are mentioned in my comments [here](https://forums.solar2d.com/t/audio-patch-to-fix-the-crash-caused-by-rtt-almixersoundcompletionevent/356162/3).

The comment [here](https://github.com/coronalabs/corona/blob/master/librtt/Rtt_PlatformOpenALPlayer.cpp#L1089), right before the task was issued, was on the right track. The `Append()` is being performed in the audio thread but the `Run()` is happening in the main thread. Since there was no synchronization an `Append()` would occasionally happen _during_ a `Run()`.

This adds a couple mutexes to guard against this. (`Rtt_Scheduler` seems to be a singleton, so I made them `static` rather than add `<mutex>` in the header file, but they could just as well be a `static` or even regular member.)

Since you _could_ have patterns like in the test, I added `Task`s to a list and synchronized them at certain points. That way the `Run()` shouldn't be stalling `Append()`s repeatedly. One possible issue is that `Delete()` could miss between extracting the list and synchronizing it, due to the two mutexes, but I didn't see any evidence of its use in the codebase. (Looks like apart from playing sounds, only the shell executing at startup and audio recording use the scheduler.) I _think_ the sync in the destructor will account for any last-minute additions, since it happens in the same thread as `Run()`.


Shchvova (approved):
seems legit. Lets' hope it won't cause wake locks/ANRs

ggcrunchy:
I've [converted the scheduler](https://github.com/ggcrunchy/corona/commit/73d88e954c48db9c9828adef3b44e20d24418259.diff) to use atomic variables rather than the `mutex`s as it was doing. (I do wonder if the scheduler might balk if `Append()` is called on the same thread as `Run()`.) Seems perfectly stable on Windows so far.

If this is needed sooner one of us can make a PR.

There are other issues: the Lua state being used on the audio thread, or rather the smart pointer to it, is not safe at all. The cleanup of `arrayOfChannelToLuaCallbacks` is also not secure. I will try to get on these pretty quickly, once I nail down the Mac relaunch behavior.

PR #660, opened 2023-12-07
Fix: always reset static profiling IDs

ggcrunchy:
As mentioned by @naveen-pcs in #645, there was a new crash with this sort of error in 3701:

`[split_config.arm64_v8a.apk!libcorona.so] Rtt_Profiling.cpp - Rtt::Profiling::EntryRAII::EntryRAII(Rtt::ProfilingState&, int&, char const*)`

This seemed to be consistent with a profile entry not being allocated, as if the ID were invalid in that part of the code. This was odd, since when the ID was `None` it just adds a slot to the array, the code seems to generally just work otherwise, and the device reports didn't seem to indicate any memory problems.

By a total fluke&mdash;while trying to figure out something else&mdash;I stumbled on [this comment](https://github.com/coronalabs/corona/blob/master/platform/android/ndk/jni/JavaToNativeBridge.cpp#L281), and in particular **niko20**'s point [here](http://groups.google.com/group/android-ndk/browse_thread/thread/3e0661379875c7ca).

That sounds right on target. These `static` IDs are new with the recent changes, and if they survived a new launch they would refer to bogus spots in the profiling list.

I had logic to update one "simulator run" value in the profiling state, if it was in simulator, that would make IDs `None` on first use, since the simulator run would differ from a "master" value. I've now changed it to just do that unconditionally.

It actually looks like the "master" value would be more generally useful (for instance in some audio fixes, rather than comparing pointers across threads), so I moved it out of the profiling state into `Runtime`, as `GetGenerationID()`.


ggcrunchy:
After looking into an error that arose with build 3701 I see that Mac's relaunch mechanics are hard to square with `static` variables at all. While away from the computer I realized this would be pretty easy to refactor and I now have something that seems to work just fine on Mac.

In light of that Mac behavior I might have to reconsider the aforementioned generation ID as well.

PR #661, opened 2023-12-08
Fix: refactor profiling IDs from statics to state members

ggcrunchy:
As mentioned in the closing comment of #660, I realized the case with `static` IDs was pretty hopeless. Thankfully, it was actually rather easy at this point to move them into the `ProfilingState`, which was on hand when they were needed. Previous design was a bit of inertia from the former stack trace-stifling version. 😄

I reproduced the issue on Mac, mentioned in #help-and-support on Discord, where launching from, say, the file dialog, would lead to a crash. With this it goes away.


Shchvova (approved):

Shchvova (approved):

ggcrunchy:
I let leak slip into the PR commit (details in 894b3ad's comments). That should be fixed now; I let this test churn through a few million objects and memory leveled off early and held steady:

```
local ii=0
local jj=0
local kk=0

Runtime:addEventListener("enterFrame", function()
  ii=ii+1
  for _ = 1, 50 do
    local r = display.newRect(1, 2, 3, 4)
    r:addEventListener("buffalo", function()
      jj=jj+1
    end)
    function r.cow ()
      kk=kk-2
    end
    r:addEventListener("cow")
    r:dispatchEvent{ name = "buffalo" }
    r:dispatchEvent{ name = "cow" }
    r:removeSelf()
  end
  if ii == 500 then
    print("!!", jj, kk)
    ii=0
  end
end)
```

I did a bit of reorganization while I was at it. I'm double-checking with those folks that I know did anything with the results so far, but made a very slight change to the output of `display.getTiming()`: when `time` is `false`, if the `what` string starts with `@` it's a "keyword" (so far either "@tableListeners" or "@functionListeners"). My "basic example" in #551 then has the following change:

```
local function PrintList (name, indent)
  local timings = table.remove(tstack) or {}
  local n = display.getTimings(timings, name)

  indent = indent or " "

  print(indent .. "> " .. name)

  for i = 1, n, 2 do
    local what, time = timings[i], timings[i + 1]

    if time then
      print(indent .. " ", what, time / 1e3)
    elseif what:starts("@") then -- keyword?
      if what:ends("Listeners") then
        print(indent .. "  *" .. what:sub(2))
      end
    else -- sublist?
      PrintList(what, indent .. "  ")
    end
  end

  tstack[#tstack + 1] = timings
end
```

ggcrunchy:
849f318's comment should read "assumption of **aligned** strings". Anyhow, with that revision it seems okay on Mac, and I'll put it through its Android paces shortly.

PR #679, opened 2024-03-01
Merging experimental branch

ggcrunchy:
Well, here goes nothing. 😄 

(A few conflict fix PRs will definitely follow.)

---

About 80 of these files are just Vulkan headers and a few binaries.

---

There are 10 or so JSON files related to tvos templates. I'm not sure what introduced them, or if they're a problem.

---

Otherwise, commentary may be found in the various PRs that have been merged into `experimental`.


platform/linux/CMakeList.txt
HBL001, line 1: Makelist has an update here ${CORONA_ROOT}/librtt/Corona/CoronaObjects.cpp, bbbbbbut... there is no such update in the ~\corona\librtt\Corona directory.  Not checked if this is specific to Linux builds.

image

ggcrunchy: Hi. If it is just a botched project file&mdash;bound to happen sooner or later, especially since I don't have a Linux testing setup yet&mdash;I'd first try just removing that file from **master**.

(However, I wouldn't say no to some real-world Linux testing of **experimental**. 😃)

Was hoping to finally get this merged properly but every time that looks promising there seem to be new iOS or Android requirements to address. 😂  (And a new conflict to fix, apparently.)

Shchvova (approved):

PR #682, opened 2024-03-12
Android: avoid main thread io and mmap the resource.car

clang-clang-clang:
This PR removes a main thread IO to avoid ANR; allows Android to use mmap to load Lua code packages; and an additional Rtt_ASSERT correction.

# Remove a main thread IO

There is a risk of ANR when `loadExpansionFiles()` in the main UI thread, so the main UI thread reading is changed to the main thread modifying the state, and the GLThread is used to read.

# Android to use mmap to load Lua code packages

This PR allows Android to load the Lua code package (resource.car) like other platforms, using mmap, to reduce runtime memory by extracting the resource.car at first launch. It doesn't need to be uncompressed every launch, so it also positively affects startup time.

During the public beta phase, it was observed that some models showed the situation where `AssetManager` prematurely returns -1 when decompressing resources (approximately 0.3%). This issue was resolved by improving the accuracy, reliability, and usability of the `extractAssetFile()` method (at least this exception is no longer observed online). The worstcase scenario here is the same as before, when the `getBytesFromFile()` method failed to load the code package into memory all at once, resulting in a black screen after startup, but usually, it will not happen again when restarting the App.

# Rtt_ASSERT correction

Additionally, while debugging other issues, I found an `Rtt_ASSERT` that could not be false. After correcting it to match the context, I guess it's not allowed to move `DisplayObject` from the `Orphanage()` group to the `Orphanage()` group again, that is, to `removeSelf` repeatedly.

Any suggestions are welcome. Thank you.

### DIFF
--- a/librtt/Rtt_Archive.cpp
+++ b/librtt/Rtt_Archive.cpp
@@ -1,6 +1,7 @@
 //////////////////////////////////////////////////////////////////////////////
 //
-// This file is part of the Corona game engine.
+// This file is part of the Solar2D game engine.
+// With contributions from Dianchu Technology
 // For overview and more information on licensing please refer to README.md 
 // Home page: https://github.com/coronalabs/corona
 // Contact: support@coronalabs.com
@@ -31,6 +32,7 @@
 	static const unsigned S_IWUSR = _S_IWRITE;    ///< write by user
 #elif defined( Rtt_ANDROID_ENV )
 	#include "NativeToJavaBridge.h"
+	#include <sys/mman.h>
 #else
 	#include <stdlib.h>
 	#include <unistd.h>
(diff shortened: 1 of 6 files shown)


Shchvova (approved):

Shchvova:
This is awesome. Earlier versions of Android had issues with mmap. I see that you have it in the wild and it seems to work now?

clang-clang-clang:
mmap works like a charm.

> Earlier versions of Android had issues with mmap.

Are there specific problem records for early versions? We have covered the API 15, which is the minimum supported by Solar2D. Since the launch of mmap, we have never observed mmap errors in the location of loading `resource.car`, but it may not be sufficient to reflect compatibility issues.

PR #699, opened 2024-04-30
[Web] Frame Rate and UI fix

cemalgnlts:
Hi,

I think a few frames are omitted because of the floating point numbers in this line.
https://github.com/coronalabs/corona/blob/ec4935044ede59342b289f4cbfbb08b5934b6e08/platform/emscripten/Rtt_EmscriptenContext.cpp#L67-L77

I can show it with JS (you can paste it into the console in Developer Options and see the result):
```js
function runTest(fps) {
  let last = 0;
  let actualFrame = 0;
  let s2dFrame = 0;
  
  const FPS = fps ?? 60;
  const MS_PER_SEC = 1000 / FPS;
  
  function onFrame(now) {
      actualFrame += 1;
  
      if(now - last >= MS_PER_SEC) {
          last = now;
          s2dFrame += 1;
      }
  
      if(actualFrame % (FPS * 3) === 0) { // run 3 seconds
          completed();
      } else {
          requestAnimationFrame(onFrame);
      }
  }
  
  requestAnimationFrame(onFrame);
  
  function completed() {
      console.log("Frame got: %s, S2D got: %s, Missed Frame: %s", actualFrame, s2dFrame, actualFrame - s2dFrame);
  }
}
```

The results I get when I run this example:
```js
runTest() // Frame got: 180, S2D got: 92, Missed Frame: 88
runTest(30) // Frame got: 90, S2D got: 31, Missed Frame: 59
```

If we update it this way and try again:
```diff
- const MS_PER_SEC = 1000 / FPS;
+ const MS_PER_SEC = 1 / FPS;
```

```js
runTest() // Frame got: 180, S2D got: 180, Missed Frame: 0
runTest(30) // Frame got: 90, S2D got: 90, Missed Frame: 0
```

I also fixed that if the status text is blank on the loading screen, the height is reset and the logo slides down.

_The commit history may look a bit dirty because the code is not compiled properly in linux (maybe I couldn't do it)._


Shchvova:
Was this fixed by other PR already?

PR #731, opened 2024-10-09
iOS: Fixes to Open Url for iOS 18

scottrules44:
" BUG IN CLIENT OF UIKIT: The caller of UIApplication.openURL(_:) needs to migrate to the non-deprecated UIApplication.open(_:options:completionHandler:). Force returning false (NO)."

Based on forums
https://forums.solar2d.com/t/on-ios-18-system-openurl-url-does-not-work/356845/2

### DIFF
--- a/platform/iphone/Rtt_IPhonePlatform.mm
+++ b/platform/iphone/Rtt_IPhonePlatform.mm
@@ -439,7 +439,8 @@ IPhonePlatform::OpenURL( const char* url ) const
 		NSURL* urlPlatform = [NSURL URLWithString:[NSString stringWithUTF8String:url]];
 		if ( [sharedApplication canOpenURL:urlPlatform] )
 		{
-			result = Rtt_VERIFY( [sharedApplication openURL:urlPlatform] );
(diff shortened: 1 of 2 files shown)


Shchvova (approved):
Lgtm. Please remove verify.

platform/iphone/Rtt_IPhonePlatform.mm
```diff
@@ -439,7 +439,8 @@ - (void)touchesCancelled:(NSSet *)touches withEvent:(UIEvent *)event
 		NSURL* urlPlatform = [NSURL URLWithString:[NSString stringWithUTF8String:url]];
 		if ( [sharedApplication canOpenURL:urlPlatform] )
 		{
-			result = Rtt_VERIFY( [sharedApplication openURL:urlPlatform] );
+			result = Rtt_VERIFY( true ); // We set to true due the fact we don't know if web link opened
```
Shchvova, line 442:
```suggestion
			result = true; // We set to true due the fact we don't know if web link opened
```

PR #737, opened 2024-10-29
Android: Save Photos to Library on Android 10+

scottrules44:
https://forums.solar2d.com/t/saving-images-to-the-gallery-hasn-t-been-working-since-android-10-display-capturebounds/356924

Android seems to lock out writing directly to Camera Roll on 10+


AleCGames:
Thank you @scottrules44 , but I need to correct you on something: the issue doesn’t occur starting from Android 13 onwards, but rather from Android 10.

scottrules44:
This has been corrected

Shchvova (approved):

PR #758, opened 2025-01-16
Fix meshes not centering when created using buffers

depilz:
Fixes this:

https://github.com/user-attachments/assets/d67928a8-0d9a-4b17-902f-99d170a1b375

Result:

https://github.com/user-attachments/assets/a09b9b98-2894-4a2d-82df-9b9ee4c77543

@ggcrunchy does this looks good to you?

### DIFF
--- a/librtt/Display/Rtt_ShapeAdapterMesh.cpp
+++ b/librtt/Display/Rtt_ShapeAdapterMesh.cpp
@@ -111,6 +111,7 @@ static const unsigned char* IssueWarning( lua_State *L, const unsigned char* buf
 static const unsigned char* GetBuffer( lua_State *L, size_t valueSize, size_t element, U32 &n, size_t &stride, const U32* outputLength = NULL )
 {
 	size_t len;
+    stride = 0;
 
 	const unsigned char* buffer = AuxGetBuffer( L, valueSize, len, n );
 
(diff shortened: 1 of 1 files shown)


ggcrunchy:
Definitely looks better in your example. 😄

Shchvova:
Reason I haven't merged it yet is because of backward compatibility. Do you think it'll break existing apps?

depilz:
Actually both of them need the center mesh code, I took it from the non-fromVertices and now it's being applied to both cases (pure Lua and buffered)

This is an initialization call, so centering the mesh will always be required.

ggcrunchy:
@Shchvova I was looking over some of the code with @depilz and it seems like the vertex offset calculation is almost replicated in the tessellator's `Update()` method; I think the only difference is that the latter is done on the already-offseted vertices.

This doesn't affect the correctness, but seems like a legitimate optimization&mdash;meshes can have plenty of points, after all. Off-hand I didn't see if the situation would _always_ occur or if the rect would want to be an optional input to `Update()`. (I could probably supply some some rough code in the next day or two.)

depilz:
@Shchvova 20ee14ab522b538dd749f730993aaf61aa4b8d5f will prevent unnecessary updates when using regular tables to update a mesh.

depilz:
Testing branch using lua tables to update meshes.

https://github.com/user-attachments/assets/256c8caf-4afd-4224-bb44-bedd6d310d55

Testing branch using buffered memory data to update meshes.

https://github.com/user-attachments/assets/ac46104e-e291-4fea-a36b-cb77fd98e4f2

They both look equally good but buffered is definitely a great improvement in performance .

@Shchvova this branch is ready to go. Can we merge it in?

Shchvova (approved):

PR #777, opened 2025-02-07
Maintenance

scottrules44:
Guess this a weird Samsung thing (made Android 14+ does not cause any issue on non-samsung device's in my testing for Android 15) Also cleaned up libplayer being added on TvOS Native Plugin Collection


Shchvova:
lucky number #777

PR #785, opened 2025-03-01
iOS: Support for Extensions

scottrules44:
Adding the ability for iOS Apps to support appx files in their apps. This could be used to support any of the following extensions via the plugin system:
Widgets, Intents, iMessage, etc

https://developer.apple.com/app-extensions/

Note that my main use for this at the moment is to support Rich Notification Extensions for an upcoming Personal Plugin for OneSignal for Corona Sim.

I will write some basic docs for this in Corona Native/ Corona Sim guides for how to support this.

I updated the signing script to sign the Appx/Extensions if there are any plugins; include them appropriately. 

Thanks,
Scott

### DIFF
--- a/platform/resources/iPhonePackageApp.lua
+++ b/platform/resources/iPhonePackageApp.lua
@@ -320,52 +320,19 @@ end
 
 --------------------------------------------------------------------------------
 
-local function getCodesignScript( entitlements, path, identity, developerBase )
-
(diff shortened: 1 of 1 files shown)


scottrules44:
Screenshot 2025-03-01 at 4 31 51 PM

Shchvova:
Nice!

Shchvova (approved):

PR #799, opened 2025-04-24
Maintenance

scottrules44:
Added support for Xcode 16.3

Note Apple has requires builds with Xcode 16
https://developer.apple.com/news/upcoming-requirements/?id=02212025a


Shchvova (approved):
lgtm. Merging in test branch... For testing

PR #800, opened 2025-04-24
Maintenance

scottrules44:
Looks like Corona Cards need to be bumped up to MacOS 15


Shchvova:
merting to test branch

Issue #794, opened 2025-04-05
Android 15 warnings displayed in the Google Play Console


Geams:
After uploading my app to Google Play, warnings are displayed.
Normally I would just ignore them, but I fear that my app will be hidden from users because of that.
### **Edge-to-edge may not display for all users**
From Android 15, all apps targeting SDK 35 will display edge-to-edge by default. Apps targeting SDK 35 should handle insets to make sure that their app displays correctly on Android 15 and later. We recommend investigating this issue and allowing time to test edge-to-edge on your app and make the required updates to your code. Alternatively, call `enableEdgeToEdge() `for Kotlin or `EdgeToEdge.enable()` for Java for backward compatibility.
### Your app uses deprecated APIs or parameters for edge-to-edge
One or more of the APIs you use or parameters that you set for edge-to-edge and window display have been deprecated in Android 15. Your app uses the following deprecated APIs or parameters:

`android.view.Window.setNavigationBarColor`
`LAYOUT_IN_DISPLAY_CUTOUT_MODE_SHORT_EDGES`
These start in the following places:

`com.ansca.corona.CoronaActivity$5.run`
`com.ansca.corona.CoronaActivity.setStatusBarMode`
To prepare for the change to Android 15, migrate away from these APIs or parameters.
### Remove resizability and orientation restrictions in your game to support large screen devices
Your game doesn't support all display configurations, and uses resizability and orientation restrictions that may lead to layout issues for your users.

We detected the following resizability and orientation restrictions in your game:
```
<activity android:name="com.ansca.corona.CameraActivity" android:screenOrientation="PORTRAIT" />
<activity android:name="com.ansca.corona.CoronaActivity" android:screenOrientation="PORTRAIT" />
<activity android:name="com.ansca.corona.CoronaActivity" android:resizeableActivity="false" />
```
`com.ansca.corona.purchasing.StoreActivity.onCreate`
To improve the user experience of your game, remove these restrictions and check that your game layouts work on various screen sizes and orientations.

appfamily:
@Shchvova @scottrules44 Is this covered in #810 ?

kan6868:
With the last issue in report.
For android device api >= 36 and sw ≥ 600dp

We need to remove screenOrientation and enable resizeableActivity = true

https://developer.android.com/develop/ui/compose/layouts/adaptive/app-orientation-aspect-ratio-resizability?hl=en#changes

PR #810, opened 2025-06-03
Android: API 35 support and 16KB support

scottrules44:
I followed instructions and tested with this script
https://cs.android.com/android/platform/superproject/main/+/main:system/extras/tools/check_elf_alignment.sh

Updated all core libs, we will need to looking any plugins to make sure they support 16kb page size
https://developer.android.com/guide/practices/page-sizes#elf-alignment

I also bumped up SDK android 15

Tested with release here
https://github.com/scottrules44/Solar2D-98/releases/tag/3723


Shchvova (approved):

appfamily:
Does this address issue #794 , the various warnings in the dev console for display changes?
image

PR #812, opened 2025-06-06
Maintenance

scottrules44:
Fixes to pull request "Core: Fixing unpack with alignment"

https://github.com/scottrules44/Solar2D-98/actions/runs/15482662331

### DIFF
--- a/librtt/Renderer/Rtt_GLTexture.cpp
+++ b/librtt/Renderer/Rtt_GLTexture.cpp
@@ -1,7 +1,7 @@
 //////////////////////////////////////////////////////////////////////////////
 //
 // This file is part of the Corona game engine.
-// For overview and more information on licensing please refer to README.md 
+// For overview and more information on licensing please refer to README.md
 // Home page: https://github.com/coronalabs/corona
 // Contact: support@coronalabs.com
 //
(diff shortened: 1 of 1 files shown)


librtt/Renderer/Rtt_GLTexture.cpp
```diff
@@ -115,9 +125,11 @@ GLint CalculateOptimalAlignment(U32 width, GLenum format)
             bytesPerPixel = 3;
             break;
             
-        case GL_RGBA:
-        case GL_BGRA_EXT:
```
clang-clang-clang, line 119: Hi there! I had a quick question about this PR. Was the removal of `GL_BGRA_EXT` an intentional change? From my understanding, this PR seems to be focused on `GL_ABGR_EXT` macro definitions, and `GL_BGRA_EXT` still feels relevant, even if the default outcome is similar.

clang-clang-clang:
Hey! Thanks for your contributions.

I noticed some code-unrelated formatting changes (like indentation/line breaks) in this PR. Since a new version hasn't been released yet, I wonder if we should aim to follow the LLVM "golden rule" of adopting existing styles for uniformity. Just a thought for consistency. Thank you!

Shchvova:
@clang-clang-clang I think right now code doesn't have consistent formatting unfortunatelly. I think if your editor applies some formatting that's OK to be honest. Any formatting would be better than whatever we have now. Or does it causes some issues?

clang-clang-clang:
There are no issues.
My simple idea is to reduce the cost of comprehension.
Due to historical reasons, it is not recommended to change the existing format unnecessarily. However, there is no problem with the new code.

Shchvova:
Agree. But also, easily mitigated with 
image

clang-clang-clang:
Yep, basically all Git tools have this option. Because I have encountered situations where blank characters affected the operation, I usually perform binary comparison and pay attention to any potential impacts. Especially, Solar2D run and be maintained on multiple platforms, across different programming languages and natural languages, and some resource contents need to be mindful of format changes.

Overall, there is no problem. As long as the team can understand each other, everything will be fine. ;D
Thank you very much for your explanation.

scottrules44:
My mistake I used Xcode to edit your code and which may have messed it up a bit.
As far as white space, my understanding is the that cpp is not really affected by white space with a small few exception. Anyway next time we look at this we can try to clean it up sometime in the future. As a side note please test on a platforms before you make any changes :)

PR #816, opened 2025-06-29
[Linux] Fix bad behavior when changing skin

kan6868:
- Skin selection dialog does not close after selecting another skin.
- Orientation is not correct when selecting skin (default is always portrait).


Shchvova (approved):

Shchvova:
Thank you so much!

PR #823, opened 2025-07-10
Core/Physics: Add WheelJoint and PistonJoint version 2.

kan6868:
To keep backward compatible with Box2D v2.4.2 upgrades, I've created an additional v2 for joints.

Need to add b2WheelJointV2 and b2PrismaticJoint to build list

```lua
local wheelJoint = physics.newJoint("wheelV2", bodyA, bodyB, anchor_x, anchor_y, axisX, axisY);

local pistonJoint = physics.newJoint( "pistonV2", bodyA, bodyB, anchor_x, anchor_y, axisX, axisY )
```

### DIFF
--- a/librtt/Rtt_LuaLibPhysics.cpp
+++ b/librtt/Rtt_LuaLibPhysics.cpp
@@ -1041,10 +1041,11 @@ InitializeFixtureFromLua( lua_State *L,
 static const char kDistanceJointType[] = "distance";
 static const char kPivotJointType[] = "pivot";
 static const char kPistonJointType[] = "piston";
+static const char kPistonJointV2Type[] = "pistonV2";
 static const char kFrictionJointType[] = "friction";
 static const char kWeldJointType[] = "weld"; // note: has no type-specific methods
 static const char kWheelJointType[] = "wheel"; // combines a piston and a pivot joint, like a wheel on a shock absorber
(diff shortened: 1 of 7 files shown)


clang-clang-clang:
Has Solar2D officially upgraded to Box2D V3? If not, wouldn't it be better to keep the same keyword, and automatically determine by the organization of the parameter such as table or position, without introducing an extra key? Just like [newText](https://docs.coronalabs.com/api/library/display/newText.html)?

kan6868:
> Solar2D đã chính thức nâng cấp lên Box2D V3 chưa? Nếu chưa, liệu có tốt hơn không nếu giữ nguyên từ khóa và tự động xác định theo cách tổ chức tham số như bảng hoặc vị trí, mà không cần thêm khóa bổ sung? Giống như [newText](https://docs.coronalabs.com/api/library/display/newText.html) ?



> Has Solar2D officially upgraded to Box2D V3? If not, wouldn't it be better to keep the same keyword, and automatically determine by the organization of the parameter such as table or position, without introducing an extra key? Just like [newText](https://docs.coronalabs.com/api/library/display/newText.html)?

No this is just an upgrade of the missing features of the current Solar2D to 2.4.2 (the last version of box2d c++), upgrading to v3 requires a rebuild from scratch for the current physics api.

My roadmap for the Solar2D physics api is to customize it as a separate version of box2d. This may include porting the new box2D v3 features into the current version.

I want to separate it from the old wheel/piston, as it will be completely different in how it works. (unlike text which is just an addition of parameters).

Issue #822, opened 2025-07-09
[Android]: SafeArea does not exclude the buttons/gesture bar on Samsung Galaxy S22 Plus and S23 Ultra


lkogut:
When I position buttons group on bottom of screen (using getSafeAreaInsets() function insets or display.safeActualContentHeight + display.safeScreenOriginY - buttons.height) it works fine on simulator, Google Pixel 8 device and some Motorola device. But on Samsung Galaxy S22 Plus and Galaxy S23 Ultra (Android 15, oneUI 7.0) system UI i rendered on my buttons.

Google Pixel 8

<img width="1080" height="422" alt="Image" src="https://github.com/user-attachments/assets/fb689d63-f076-4184-b587-30e6c9335d52" />

Samsung Galaxy Galaxy S23 Ultra (Android 15, oneUI 7.0) or S22 Plus

<img width="746" height="202" alt="Image" src="https://github.com/user-attachments/assets/68bc5d99-eef0-4e1a-b89b-4b60f7f3b367" />

<img width="746" height="202" alt="Image" src="https://github.com/user-attachments/assets/19adba52-f16f-4255-96a9-4dd399b3af25" />

I don't have direct access to this devices to add tested sample code, but I think this example will show the problem:
https://docs.coronalabs.com/api/library/display/getSafeAreaInsets.html

lkogut:
I have just upgraded my Pixel 8 to Android 15 and now I have the sam problem with Pixel 8. So it's not S22/23 problem, but rather Android 15.

<img width="1080" height="312" alt="Image" src="https://github.com/user-attachments/assets/8527fa57-6969-4031-b73b-34e09e2025b2" />

Now I can debug it and it looks like this:

<img width="241" height="45" alt="Image" src="https://github.com/user-attachments/assets/c156bb11-b28d-4cf1-8314-0993c490ad53" />

PR #826, opened 2025-07-15
Android: Maintenance

scottrules44:
Android 35 fixes:
Fixes for safe space on bottom, address other google warnings, include updates for subrepos

Address issues:  #794, #822


clang-clang-clang:
Note that the fix for the navigation bar height, which does fix height getter on Android 15, but may break the pre-Android 15 value.

PR #832, opened 2025-08-22
Fix windows-2022 build

zero-meta:
We can ignore the **7-Zip** extraction error about "**Dangerous link**."

According to the **GitHub Actions runner** documentation, in Visual Studio 2022, the `<WindowsTargetPlatformVersion>` is set to "10.0.19041.0," and the `<PlatformToolset>` is set to v143.

The [**redistributable DLLs**](https://github.com/coronalabs/binary-data/releases/download/1.0/redist.win32.7z) should be upgraded; otherwise, the built program will not run properly. Here is my [upgraded version](https://github.com/zero-meta/solar2d_binary_data/releases/download/v1.1/redist.win32.7z). 

You can change code:
`curl -sL https://github.com/coronalabs/binary-data/releases/download/1.0/redist.win32.7z -o redist.win32.7z`
in **build.yml** to download my version or upgrade yourself.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -480,6 +480,7 @@ jobs:
         run: |
           7z x -snld SourceCode/corona.tgz
           7z x -snld corona.tar
+        continue-on-error: true
       - run: ./tools/GHAction/daily_env.sh
         shell: bash
       - run: mkdir -f docs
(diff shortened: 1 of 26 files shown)


Shchvova:
I already have a or for this, current problem is that iOS isn't building :(

PR #848, opened 2025-09-25
Further fixes for time transform snags; removed need for isTimeDependent; time-based builtin effects now using transforms

ggcrunchy:
(This probably needs to wait for word that the Vulkan backend behaves right, beyond just compiling, but otherwise feels good.)

---

In Discord it came up that time transforms still have a few issues. Assigning one to a shader was broken. Adding **isTimeDependent** to a shader when it wasn't used could cause some things to break, as well.

It was also mentioned that the marching ants builtin (a time-based effect) got worse over time.

---

The broken shader assignment was something silly: doing a `lua_isstring()`, when the string in question had already been popped off the stack, rather than check for the non-`NULL` result.

The fixed behavior is used by the time-based builtins, described below.

---

**isTimeDependent**:

Once a scene has been drawn, Solar will leave it alone if nothing changes, i.e. no new objects; objects being moved or rotated; etc. This is mostly the right behavior, but time-based effects _do_ need to be redrawn. The **isTimeDependent** flag, passed to shader definitions, was used to convey this information.

The shader author needed to remember to add this flag. And to be careful to remove it / not add it otherwise, apparently.

The time transforms could also use this information as a hint about when they needed to be called. I was attempting some caching here (from #511), and it apparently blew up if you wrongly provided the flag.

---

This PR ignores **isTimeDependent** entirely and discovers the dependency in a different way.

We still query `ShaderResource::UsesTime()`, but the state itself is being found automatically: if the `CoronaTotalTime` uniform is bound, following a `GLProgram::Create()`, the flag is set. (Any variant, e.g. # of masks or distorted rects, can be responsible, and the flag is taken as normative for all of them.) This is within the first `GLProgram::Bind()` of the variant; there's an untested `DEFER` path too, if anything ever wants to use that.

There are corresponding Vulkan backend parallels, too.

The scene will have already been updated by the time the `Create()` is performed, and thus be unaware of any new time dependency. If any `ShaderResource` had its flag set, the scene is therefore invalidated; it will pick up on anything new the next time around.

Some commentary about this may be found in **Rtt_Scene.cpp**.

The default time transform is applied to the raw total time when the scene begins rendering, and saved in a value. This is used by shaders without custom transforms; otherwise the raw time is provided, so that it can be transformed if necessary. In the latter case the `CoronaTotalTime` uniform's update is couched between bind and unbind commands for the transform.

Other than the default up front, no attempt is now made at transform caching, or even checking that the shader _uses_ time, while issuing commands. If the time uniform is actually bound, any transform is resolved when the corresponding command is executed.

---

There are three time-based builtins: **generator.marchingAnts**, **filter.wobble**, and **generator.random**.

Each was unbounded and so would eventually suffer some artifacts. However, all three are "modulo"-friendly cases and were given such a transform.

Marching ants is a repeating pattern. It moves at a certain speed (16), and once it's covered two bars' worth of distance (off and on, so 6 + 6), we have an identical-looking effect. Thus its range is (6 + 6) / 16 = .75 seconds.

Wobble uses time inside a sine. One period will have elapsed when 3 * t = 2 * pi, so our range is 2 / 3 * pi.

Random was just doing a `fract()`, so there the range is 1.

These numbers are found in the effects themselves. I added similar commentary in there. Old code is still present, commented out; forgot about it during cleanup, but I guess it's fine for comparison purposes.

---

I was testing with this:

[fx_tests.zip](https://github.com/user-attachments/files/22531100/fx_tests.zip)

I had it set up to try the time-based builtins as well as copies of their code, with minor modifications in how they calculated time, and now without using **isTimeDependent** at all. (Those versions might be a little odd themselves over a certain duration, since they should now use the default "pingpong" behavior, but were used to manually check the modulo ranges and the code still reflects that.)

Some of the `if true`s and such were for quickly disabling certain effects to make sure we didn't still have **isTimeDependent** at work, as well as to verify that a delay between shader creation and bind would still behave right. (This is currently in place and the reason the effects take a few moments to start.)


Shchvova (approved):
This is very nice. I'm worried about backwards compatibility. 
It wouldn't affect existing code, would it?

ggcrunchy:
Any incompatibility would indicate a bug, as far as I can tell. It shouldn't stomp on any existing expected behavior. Also, there were only those three builtins, and I confirmed they were using the transforms.

marksolar2d:
Tested this on a new laptop and you just broke 10+ of my shaders. It stomps all over things for sure!  Reverted to 3722 and everything works.... You could at least asked me to test it Steve - knowing I am shader heavy!  Admittedly I haven't updated because things on 3722 were stable, but still this is still a massive breaking change! Have PM'd you a shader you can test with.

PR #850, opened 2025-10-09
Fix incorrect extraction of the get-task-allow flag from the provisioning profile

depilz:
Fixes an invalid provisioning profile issue I have been facing since I upgraded to Tahoe.

https://forums.solar2d.com/t/ios-deploy-fails-after-macos-upgrade-a-valid-provisioning-profile-for-this-executable-was-not-found/357564

I get this log from the Engine when building with `debugBuildProcess = true`
```get_task_allow_setting:     "get-task-allow" => true```

When the expected is just:
```get_task_allow_setting:     true```

I have been able to resign my app properly with that command, but I haven't been able to test this change from the engine itself.

### DIFF
--- a/platform/resources/iPhonePackageApp.lua
+++ b/platform/resources/iPhonePackageApp.lua
@@ -518,14 +518,8 @@ local function generateXcent( options )
 		print("get_task_allow_setting: ".. tostring(get_task_allow_setting))
 	end
 	if get_task_allow_setting ~= "" then
-		-- set the value appropriately
-		if get_task_allow_setting == "true" then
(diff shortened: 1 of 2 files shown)


bulupe:
Thank you, with 3725 I can build but I can not copy to device, it says no device found although I can see the device and I can install the app via Devices and Simulators in Xcode.

There is one more thing happening with iOS, it seems we can not see print results in Console anymore on device, I thought it is because it is downloaded from store but even development builds the situation is the same.

Screenshot 2025-10-17 at 12 42 54

Sorry to bother you but it makes really hard to debug on device, like ads or ump.

Thanks and Best Regards

PR #868, opened 2025-11-13
Core: fix iOS highp misjudge by set a default value without relying on a highp check

### LINKED ISSUE #795: iOS - object.fill.effect = "filter.hue" doesn't work anymore
Because of Apple's upcoming deprecation of the iOS 17 SDK, I've made the jump from Solar2D 3699 building with Xcode 15.0.1 to Solar2D 3718 building with 16.2. Everything builds fine, but when I test my app, object.fill.effect = "filter.hue" doesn't seem to be working any more.

The image below demonstrates the difference between builds made with Solar2D 3699 and 3718.

Image

clang-clang-clang:
This fix https://github.com/coronalabs/corona/issues/795.

The changes in b951bf5d657ab7047ce0fec451b5d6ea3260bac8 added `TimeTransform` to `DisplayDefaults`, which accesses `GetGpuSupportsHighPrecisionFragmentShaders()` to configure default `TimeTransform`.
https://github.com/coronalabs/corona/blob/f09b81447900f8e89cb8a7eb5a0db7dc5762f691/librtt/Renderer/Rtt_GLCommandBuffer.cpp#L233

AFAIK, due to differing gl initialization flows between the Apple platform and Android (less familiar with other platforms), the former's setup may occurs at
https://github.com/coronalabs/corona/blob/f09b81447900f8e89cb8a7eb5a0db7dc5762f691/platform/iphone/CoronaViewPrivate.mm#L564.
In contrast, Android's gl initialization is completed on the Java side, prior to the C/C++ Runtime initialization. The Apple platform, potentially due to context not been setup, may then access incorrect values.
This explains why Android is unaffected, while iOS misjudged highp support, so this PR addresses the issue by deferring the `TimeTransform` initialization of `DisplayDefaults`.

I'm not entirely familiar with the recent changes to TimeTransform, but I believe lazy loading could be one solution. Alternatively, perhaps the functionality of Display itself shouldn't be called within DisplayDefaults, which is part of the Display construction flow. In other words, is it possible to set a default value without relying on a highp check? 
Hi, @ggcrunchy, please investigate if there's a more suitable approach? Thank you.

### DIFF
--- a/librtt/Display/Rtt_Display.cpp
+++ b/librtt/Display/Rtt_Display.cpp
@@ -308,6 +308,8 @@ Display::Initialize( lua_State *L, int configIndex, DeviceOrientation::Type orie
         result = true;
 
 		fShaderFactory = Rtt_NEW( allocator, ShaderFactory( *this, programHeader, backend ) );
+
+        fDefaults->SetTimeTransform(nullptr);
 	}
 
     return result;
--- a/librtt/Display/Rtt_Display.cpp
+++ b/librtt/Display/Rtt_Display.cpp
@@ -308,8 +308,6 @@ Display::Initialize( lua_State *L, int configIndex, DeviceOrientation::Type orie
         result = true;
 
 		fShaderFactory = Rtt_NEW( allocator, ShaderFactory( *this, programHeader, backend ) );
-
-        fDefaults->SetTimeTransform(nullptr);
 	}
 
     return result;
--- a/librtt/Display/Rtt_DisplayDefaults.cpp
+++ b/librtt/Display/Rtt_DisplayDefaults.cpp
@@ -67,7 +67,7 @@ DisplayDefaults::DisplayDefaults()
 
 	fTimeTransform = &sTransform;
 
-//	SetTimeTransform( NULL );
+	SetTimeTransform( NULL );
 }
 
 const TimeTransform* DisplayDefaults::GetTimeTransform() const
--- a/librtt/Display/Rtt_DisplayDefaults.cpp
+++ b/librtt/Display/Rtt_DisplayDefaults.cpp
@@ -67,7 +67,7 @@ DisplayDefaults::DisplayDefaults()
 
 	fTimeTransform = &sTransform;
 
-	SetTimeTransform( NULL );
+//	SetTimeTransform( NULL );
 }
 
 const TimeTransform* DisplayDefaults::GetTimeTransform() const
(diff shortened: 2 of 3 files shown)


clang-clang-clang:
The primary concern is that the highp check is a one-time operation that gets cached. If it's performed too early, it might not be able to retrieve the correct value. Therefore, would it be better to bypass the highp check and provide a default fallback value (applicable across all precisions, perhaps 1 in earlier versions)?

clang-clang-clang:
> 😄 Just use a "safe for `mediump`" value.

Got it. Thank you for your patient explanation.
So I followed the previous 50 and simply removed the highp check, and the sample [Graphics/GeneratorViewer](https://github.com/coronalabs/samples-coronasdk/tree/master/Graphics/GeneratorViewer) didn't show any problems.

I need more time to digest precision, frame rate, and jitter avoidance. @~@

Shchvova (approved):

Shchvova:
@clang-clang-clang  Do you think we should merge this?

clang-clang-clang:
> @clang-clang-clang Do you think we should merge this?

@Shchvova IMO, Yes.
Based on the context of @ggcrunchy, it's merely a default value. Those who genuinely need it will customize it. Here, we're maintaining basic functionality to ensure consistent behavior with pre-experiment versions.

ggcrunchy:
I have heard from a couple people that the short default is affecting them. I assume they were testing on a **highp** and the default pingpong / 4.5 hour thing up to that point. So I'm going to try to restore that (very slightly tuned), probably along these lines:

[Defaults.patch](https://github.com/user-attachments/files/32156061/Defaults.patch)

Up until you've stomped on the default transform (if ever) I just bake a lazily-built one into a new shader, using the "supports **highp**?" query. At this point I assume it's safe to read GL state. 😄

---

I had some ideas about being able to peg a small set of "clocks" and maybe reset them, e.g. with id 4 you'd have `graphics.resetClock(4)` and some `clockID = 4` (or whatever) in `graphics.defineEffect()`, but when I did a review I wasn't sure how to do it cleanly. That would let you make a few granularities, like the current level, the current cutscene, etc. and not one global one since startup that just keeps growing.

---

I see in the GL ES 2.0 reference card both `GL_ES` ("Defined and set to integer 1 if running on an OpenGL-ES Shading Language.") and `GL_FRAGMENT_PRECISION_HIGH` ("1 if highp is supported in the fragment language, else undefined [4.5.4]"). (The latter seems to disappear in mainline GL and ES 3.0+.) Might be some way to use that (and maybe `__VERSION__`) to auto-set `FRAGMENT_SHADER_SUPPORTS_HIGHP`, for built-in effects and maybe beyond? @Shchvova Any idea if trying it that way caused problems?

---

Anyhow, those last two things are just sort of incidental comments.

PR #891, opened 2026-03-28
fix(linux): enable Solar2DBuilder Android builds on Linux

### LINKED ISSUE #890: Community: Linux Solar2DBuilder builds for CI/CD (Android APK)
Hey Solar2D community 👋

We needed to build Android APKs from Solar2D on Linux CI runners, but there's no official Linux build of `Solar2DBuilder`. After digging into the source, it turned out to be mostly `#ifdef` gaps — Solar2D's C++ code is cross-platform, it just hadn't been wired up for Linux.

We fixed 5 small issues, bundled the result, and published it as a community repo:

**https://github.com/Cubeage/solar2d-linux**

- Pre-built binaries with bundled libs (works on any distro)
- Auto-rebuilds when Solar2D releases a new version
- Patches documented if you want to build yourself

The fixes are straightforward — adding a `Rtt_LINUX_ENV` branch wherever Mac/Win were handled, wiring up `GetResourceDirectory()` via `/proc/self/exe`, and using `$TMPDIR` instead of `/TemporaryFiles`. Would be happy to upstream these as a PR if the team is interested.

Sharing here in case it's useful to others!

shtse8:
Fixes #890

## What

Enables `Solar2DBuilder` to build Android APKs on Linux. The C++ code already had `Rtt_LINUX_ENV` defined and compiles fine — it just had Mac/Windows branches without a Linux equivalent in a few places.

## Changes (5 files)

| File | Fix |
|------|-----|
| `CMakeLists.txt` | Simplify root cmake — was duplicating platform logic that belongs in `platform/linux/CMakeList.txt` |
| `platform/linux/CMakeList.txt` | Add `Rtt_AndroidSupportTools.c` to Solar2DBuilder sources; add `CORONABUILDER_ANDROID` compile definition |
| `Rtt_AppPackagerAndroidFactory.cpp` | Add `Rtt_LINUX_ENV` branch for `AndroidValidation.lua` path (was Mac/Win only) |
| `Rtt_AppPackagerFactory.cpp` | Implement `GetResourceDirectory()` for Linux via `/proc/self/exe` readlink |
| `tools/CoronaBuilder/main.cpp` | Use `$TMPDIR` instead of hardcoded `/TemporaryFiles` (root-owned on Linux CI) |

## Tested

Successfully builds Android APKs on Ubuntu 22.04 self-hosted runners. We've been using this in production CI for our Solar2D games.

Community build repo with pre-built binaries + auto-update workflow: https://github.com/Cubeage/solar2d-linux


chkuendig:
Heads up: #936 covers this ground, opened as a supersession rather than a competing patch.

The overlap is the Linux `GetResourceDirectory()`, the `AndroidValidation.lua` path branch, and the `$TMPDIR` fix. The `$TMPDIR` one is your find — a root-owned `/TemporaryFiles` is not a failure I would have guessed at.

Three deliberate differences:

- #936 also installs `AndroidValidation.lua`. This PR adds the Linux branch that points at it, but nothing copies it into `Resources/`, so the path resolves to a file that is not there. #936 adds the `file(COPY ...)` to `platform/linux/CMakeResources.txt`.
- `GetResourceDirectory()` reuses the existing `GetStartupPath()` from `Rtt_LinuxUtils.h` rather than open-coding the `/proc/self/exe` readlink.
- #936 leaves the root `CMakeLists.txt` alone. The rewrite here drops `CORONA_ROOT`, `BUILD_NUMBER` and `YEAR`, and `platform/linux/CMakeList.txt` uses `${CORONA_ROOT}` — including for the resource copies — so it would be unset.

#936 additionally gates the HTML5 builder and fixes `display.save()` colours on Linux, which are outside what this PR set out to do.

PR #894, opened 2026-04-08
(removed)

labolado:


clang-clang-clang:
Hi, just curious, is it 6 FPS during testing or just a typo? Thanks.

```Lua
    ...
    if frame >= 360 then
        print("PASS: 60 seconds without crash")
        os.exit(0)
    end
    ...
```

labolado:
Good catch — that was leftover test infrastructure. Updated the PR with a clean minimal reproduction (7 lines for macOS + MallocScribble, full test for iOS).

PR #916, opened 2026-04-29
core: Add Edge WebView2 support with IE fallback

usmandotmughal:
This change adds **Microsoft Edge WebView2** support to the engine’s core WebBrowser implementation, enabling a modern Chromium-based backend when the WebView2 runtime is available. The system now detects runtime support at startup and uses Edge when possible, while still falling back to the existing Internet Explorer (ActiveX) implementation for compatibility on older systems. **WebView2** has been integrated through the official **Microsoft.Web.WebView2** **NuGet package**, which provides the required headers and loader components.

During integration, a naming conflict was found between Microsoft’s `winstring.h` (required by WebView2) and the engine’s existing `WinString.h`. To resolve this, the internal string implementation was renamed from `WinString` to `RttString`, ensuring compatibility with Windows Runtime APIs without conflicts.


ggcrunchy:
Looks like a LOT of local Git details get dragged in, e.g. "file mode changed" all over. (I've been hit by similar things. 😄 ) Also seems to revert some of the (recent) last merge. Probably worth doing a fresh pull of the upstream master and layering your changes on top.

PR #917, opened 2026-05-04
core: Add Edge WebView2 support with IE fallback

usmandotmughal:
This change adds **Microsoft Edge WebView2** support to the engine’s core **WebBrowser** implementation, enabling a modern Chromium-based backend when the **WebView2** runtime is available. The system now detects runtime support at startup and uses **Edge** when possible, while still falling back to the existing Internet Explorer (ActiveX) implementation for compatibility on older systems. **WebView2** has been integrated through the official **Microsoft.Web.WebView2** **NuGet package**, which provides the required headers and loader components.

During integration, a naming conflict was found between Microsoft’s **winstring.h (required by WebView2)** and the engine’s existing WinString.h. To resolve this, the internal string implementation was renamed from **WinString** to **RttString**, ensuring compatibility with Windows Runtime APIs without conflicts.


usmandotmughal:
@Shchvova, @ggcrunchy I have created a new PR for this please review.

ggcrunchy:
Well, first off, I'm just a random guy here, so I don't have any say. 😄 But I do like the actual idea and I know something like this has been requested in the past.

---

What does this add when you build, e.g. extra DLLs? Can you omit them, assuming you don't need them, and everything else still runs? (Does it gracefully fail if you _do_ need them?)

---

I do wonder if you even need the `WinString` -> `RttString` stuff. I looked through the files and it seemed like there were very few places that would introduce the clash. (I saw an `#include` in the browser C++ file.) That seems like it could be addressed with a little attention at the source level. If nothing else, put the "old" `WinString` in a namespace and do `using` in the C++ files? (Still changes a bunch, but probably less, and shouldn't really hit all the projects.)

Also, the NuGet thing showing up in all the projects is a bit of a code smell. I wonder if the WebView2 project couldn't just be built separately like many other things in `external` and then linked in? (Seems like otherwise we'd add a lot of build steps and an SDK in the common case.)

On that note, does stuff still build on older setups? That's another gotcha that might come with the SDK.

---

I think this is roughly what's come to mind so far, but will add more if I think of it. 😃

usmandotmughal:
Yeah, totally fair points.

I did try it without the extra DLLs, but the simulator wasn’t working, so I had to include them. It still falls back to the IE implementation if WebView2 isn’t available.

For the WinString → RttString change, WebView2 requires Windows Runtime headers (winstring.h), which conflicted with the existing WinString, so renaming was the most reliable fix.

Agreed on NuGet—it’s not ideal, just the quickest way to integrate.

PR #919, opened 2026-05-13
Android: Support Android 13+ predictive back gesture

zero-meta:
Enable Android 13+ predictive back gesture support. Without this, swiping back bypasses Corona’s input system entirely—developers can’t intercept the back event via Lua **key** handlers, and the app terminates instead of navigating back.


Shchvova:
Would it break existing behaviours?

zero-meta:
No impact. 
The predictive back logic is conditionally applied only on API 33+, leaving existing behavior unchanged on older versions.

PR #921, opened 2026-05-18
Android: configChanges support

scottrules44:
Have an orientation plugin, would like for to enable locking need access configChanges

### DIFF
--- a/platform/android/template/update_manifest.lua
+++ b/platform/android/template/update_manifest.lua
@@ -783,10 +783,17 @@ if (string.len(stringBuffer) > 0) then
 end
 manifestKeys.USER_SUPPORTS_SCREENS = stringBuffer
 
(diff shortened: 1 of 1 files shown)


Shchvova (approved):

Shchvova:
Plugins can have manifest in AAR archive. That would be cleaner, wouldn't it be?

Shchvova (commented):
lets see if you can do this with the AAR manifest first

Shchvova:
Some of the examples:
https://github.com/search?q=org%3Asolar2d+%22tools%3A%22&type=code

scottrules44:
@Shchvova Gotcha, good idea, trying this now :)

Shchvova:
Guess it worked?

scottrules44:
Yeah was able to just use gradle

PR #936, opened 2026-08-14
Linux: enable HTML5 and Android builds, fix display.save() and network.requestId

chkuendig:
I recently started using Solar2D in [Docker](https://github.com/chkuendig/docker-solar2d/)  (mostly for Github Action automations and to use [solar2d-mcp](https://github.com/sensiblecoder/solar2d-mcp) and noticed a few issues:

**build fixes:**
- linux: enable the HTML5 builder in Solar2DBuilder https://github.com/coronalabs/corona/pull/936/changes/2a40a997ec2202fc003443ffadd5a4f179721503
  CORONABUILDER_HTML5 was never defined for the Linux build, so Solar2DBuilder rejected html5 as an unknown platform. Defining it wires up the existing HTML5 packager that was already compiled in.
-  linux: enable Android builds in Solar2DBuilder https://github.com/coronalabs/corona/pull/936/changes/36cb5ce3f3584e71eb5732a28e67ca08ccc00fbc
Same shape: CORONABUILDER_ANDROID undefined, plus 
`AndroidValidation.lua`  wasn’t copied next to the builder, so the packager couldn’t find it at runtime. This supersedes #891
-  linux: move builder runs to a temp directory https://github.com/coronalabs/corona/pull/936/changes/7608f1f7f2bf87e40f046bea0718db865b353cfc
The four sandbox paths started as empty strings, so appending gave `/Documents`, `/TemporaryFiles`, `/CachedFiles`  and `.system`  — the filesystem root, writable only by root. Each run now gets its own `mkdtemp` directory under TMPDIR (atomic, mode 0700), removed when the run ends, matching the per-process directory the Windows builder already creates. 

**screenshot fixes**
- linux: fix the channel order of saved framebuffer captures https://github.com/coronalabs/corona/pull/936/changes/efd41a5c4ccbed0f64a241259cd96a1712858cb6
A Linux capture is tagged kBGRA, and both writers read that name as the byte order. The capture is actually A,R,G,B, so savePNG emitted G,R,A,B and saveJPG emitted G,R,A — red and green swapped, blue taken from the alpha byte. Also rejects formats no caller produces rather than misreading them, and fixes a `malloc`/`unique_ptr`-deleter mismatch.
-  linux: honour display.save()'s jpegQuality https://github.com/coronalabs/corona/pull/936/changes/edda9326e34a6beaa4112583cc124aa0e52dcd6c
`LinuxBaseBitmap::SaveBitmap` had no quality parameter, so `LinuxPlatform::SaveBitmap` discarded the caller’s value and passed a literal `75` into an API expecting 0–1 and `saveJPG` set quality before `jpeg_set_defaults`, which resets it. 
**network fixes**
- linux: fix requestId handling in the network library https://github.com/coronalabs/corona/pull/936/changes/f14cc672bd28b772c7b6d07300924e6e24abf0fe
`fRequestID` was never assigned and never read, so `event.requestId` came back nil even though `network.request() `had handed the caller an id — listeners had no way to match a response to its request. The id now lands between "url" and "bytesTransferred", the slot the Apple and Windows backends use, and the field is unsigned to match the counter. 
Separately, `network.cancel()` fed lua_tonumber straight into the request map: a missing or invalid argument all became 0 and could cancel a request the caller never named, returning true. The argument is now validated as a finite integral number in range.


chkuendig:
@Shchvova Updated this as the initial version had some issues. Let me know if there's anything to change/clarify. I have a feeling the linux path has not been been very much exercised in the past, but having it for automations (Agentic AI or just CI/CD) it's super useful.

PR #940, opened 2026-09-16
Further bugfixes for time transforms

ggcrunchy:
These have proven difficult!

The provisional fix for #868 affected some existing projects. The default time transform behavior would default to a very narrow **mediump**-ish time range, unfortunately easily observable; and while not especially great as far as correctness goes, the **highp**-ish route at least _seemed_ far better. (Time transforms are a partial attempt to address the issue underlying all of this&mdash;at need I could cite many an example outside Solar&mdash;but it's also just a deceptively hard problem. 😄)

---

**HEADS-UP**: There is a tiny amount of Vulkan fixup needed to build this on Windows; I need to reconcile some missing bits in [this](https://github.com/user-attachments/files/32276894/d5s.patch), and maybe also do some quick Windows tests to verify the results. Am posting late at night or I'd power through with it.

(I have completely forgotten what "d5s" meant. 😂)

---

I tried to summarize what gets done in the commit comments, scrounged up from a separate test branch. But anyhow, it tries to gut a lot of the caching snarls that have plagued previous attempts; _always_ has a transform in place for an effect, even when unused; and just assumes we are "using time" on the first frame an effect comes into use, despite the actual details needing to be discovered. (I have [pending work](https://github.com/ggcrunchy/corona/tree/SamplerAndShape) that will also want this better-to-seek-forgiveness workflow.)

---

There are some additional transform methods now, and you can test a couple of them by commenting out certain bits here (although this is quite a grab bag of tests!):

[interp.zip](https://github.com/user-attachments/files/32278515/interp.zip)

There were some 1D noise tests here, in particular.

---

I'll try to grind out the pending work soon, and include the remaining tests I'd done for it.


Shchvova (approved):

ggcrunchy:
Okay, added Vulkan stuff too. Got stuck on something else (sync with stdio call not found, on Windows... not sure what changed there), so haven't tried a proper build, but probably close if not there.

---

This was a test for some of the new transforms (set `func` to one of the choices in the list):

```
local func = "smoothedModulo"
local time_transform

if func == "smoothedModulo" then
  time_transform = { range = 1.5, smooth = .35 }
elseif func == "r2Noise" then
  time_transform = { range = .345 }
elseif func == "lissajousNoise" or func == "none" then
  time_transform = {}
elseif func == "modulo" then
  time_transform = {}
end

if time_transform then
  time_transform.func = func
end

graphics.defineEffect{
  category = "generator", name = "test",
  
  timeTransform = time_transform,
  
  vertex = [[
    P_POSITION vec2 VertexKernel (P_POSITION vec2 pos)
    {
      pos.y += 25. * CoronaTotalTime;
    
      return pos;
    }
  ]]
}

  Runtime:addEventListener("enterFrame", function(event)
    local r = display.newRect(display.contentCenterX, display.contentCenterY, 5, 5)
    
    r.fill.effect = "generator.custom.test"
    
    r:setFillColor(math.random(), math.random(), math.random())
    
    transition.to(r, { x = r.x - 400, alpha = 0, time = 3750, onComplete = display.remove })
  end)
```

ggcrunchy:
Missed one thing in that batch, but seems to check out now.

